PR #2, opened 2023-09-20
Map-held computed constants

liach:
This is an outline of the 2 computed-constant maps I proposed on leyden-dev list; they are copied from existing cc list and immutable map. They should be constant-foldable in theory but need to be confirmed by testing.

During implementation, I feel that the current CC-list impl can be improved by changing the CAS to CAE, so we just take the witness value instead of calling volatile get explicitly again.

I don't see where the condensers are; can probably think about how these maps can be precomputed by condensers too.

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [ ] Commit message must refer to an issue

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/2/head:pull/2` \
`$ git checkout pull/2`

`$ git checkout pull/2` \
`$ git pull https://git.openjdk.org/leyden.git pull/2/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2`

`$ git pr show -t 2`

</details>

<a href="https://git.openjdk.org/leyden/pull/2.diff">https://git.openjdk.org/leyden/pull/2.diff</a>

</details>


src/java.base/share/classes/jdk/internal/constant/IndexedComputedConstantMap.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private final IntFunction<? extends K> keyMapper;
+    private final ToIntFunction<? super K> keyUnmapper;
+    private final Function<? super K, ? extends V> provider;
+    @Stable
+    private final ComputedConstant<V>[] values;
+
+    @SuppressWarnings("unchecked")
+    public IndexedComputedConstantMap(int size,
+                                      IntFunction<? extends K> keyMapper,
```
minborg, line 53: If we want an enum map, we could just provide:
```
ICCM(Foo.values().length,
      Foo:ordinal,
      k -> e.values[k],
```

Maybe there should be a convenience method for this (that might be a bit optimized too).
      
      
```

File: src/java.base/share/classes/jdk/internal/constant/IndexedComputedConstantMap.java [hunk 2]
```diff
@@ -0,0 +1,151 @@
...
+    private static final long ARRAY_INDEX_SCALE = Unsafe.getUnsafe().arrayIndexScale(ComputedConstant[].class);
+
+    private final IntFunction<? extends K> keyMapper;
+    private final ToIntFunction<? super K> keyUnmapper;
+    private final Function<? super K, ? extends V> provider;
+    @Stable
+    private final ComputedConstant<V>[] values;
+
+    @SuppressWarnings("unchecked")
+    public IndexedComputedConstantMap(int size,
```
minborg, line 52: In theory, it would be possible to assert the invariant that the keyMapper and keyUnmapper match. However, this might defeat the purpose of the construct.

minborg:
@liach Could you please fix the merge conflicts and I will be able to merge this PR.

minborg:
Keep alive

liach:
Keep alive.

liach:
Should I close this in favor of the stable values PR in jdk?

minborg:
> Should I close this in favor of the stable values PR in jdk?

Yes. Please do that. Thanks for your effort with this PR!

rose00:
FTR, here is an example sketch of a compact but expandable container for lazy values, built on top of the VM-internal `@Stable` annotation.

https://cr.openjdk.org/~jrose/draft/StableList.java.txt

PR #7, opened 2024-03-19
Save and relocate stop msg strings in cached code on aarch64

adinn:
On aarch64 MacroAssembler method stop() plants an illegal instruction and pokes the address of its msg string (C string) in the immediately following word. This is used by the Linux and bsd signal handlers to print an error message (Windows ignores this argument). When a generated stop() is saved to and restored from the SCCache the address may not be correctly relocated (specifically, this manifests with c1_Runtime1 stub names).

The strings should be added to the SCCache and loaded (relocatably) into c_rarg0 for the handlers to find, as happens with x86.

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [ ] Commit message must refer to an issue

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/7/head:pull/7` \
`$ git checkout pull/7`

`$ git checkout pull/7` \
`$ git pull https://git.openjdk.org/leyden.git pull/7/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 7`

`$ git pr show -t 7`

</details>

<a href="https://git.openjdk.org/leyden/pull/7.diff">https://git.openjdk.org/leyden/pull/7.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -2703,8 +2703,11 @@ void MacroAssembler::resolve_global_jobject(Register value, Register tmp1, Regis
 
 void MacroAssembler::stop(const char* msg) {
   BLOCK_COMMENT(msg);
+  // load msg into r0 so we can access it from the signal handler
+  // ExternalAddress enables saving and restoring via the code cache
+  lea(c_rarg0, ExternalAddress((address) msg));
   dcps1(0xdeae);
-  emit_int64((uintptr_t)msg);
+  SCCache::add_C_string(msg);
 }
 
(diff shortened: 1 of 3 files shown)


adinn:
Testing:
This is hard to test by executing the affected code as the stop calls are only executed when something is wrong in the VM.

I verified by eyeballing loaded AOT code that the current generated code includes invalid addresses for stop messages associated with some c1_Runtime1 stubs.

I also verified by eyeballing loaded AOT code that the newly generated code installs the address into r0 using a relocatable load. The modified handler code (which now uses the value passed in r0) has not been executed.

adinn:
Closing this as I see we are not using PRs for Leyden

PR #15, opened 2024-09-06
Workaround CacheDataStore use with source launcher

shipilev:
I was trying to test the simplest workload that involves javac, and that is source launcher. When I attempt to use source launcher with `-XX:CacheDataStore`, it fails with:

```
$ rm -f app.cds*; build/linux-x86_64-server-release/images/jdk/bin/java -XX:CacheDataStore=app.cds -Xmx256m -Xms256m -XX:+UnlockExperimentalVMOptions -XX:+UseEpsilonGC HelloStream.java
[0.001s][warning][cds] optimized module handling/full module graph: disabled due to incompatible property: jdk.module.addmods=ALL-DEFAULT
Error occurred during initialization of VM
CacheDataStore cannot be created because AOTClassLinking is enabled but full module graph is disabled
```

Source launcher adds `--add-modules=ALL-DEFAULT`:

```
$ man java
...
       In source-file mode, execution proceeds as follows:

...
       • The compiled classes are executed in  the  context  of  an  unnamed  module,  as  though  --add-mod‐
         ules=ALL-DEFAULT  is  in  effect.  This is in addition to any other --add-module options that may be
         have been specified on the command line.
```

I think we can work that around specifically for Leyden, and also leave a TODO breadcrumb in the code that we need to figure this out on CDS side. This PR does that workaround. With it, source launcher works with simple examples:

```
$ build/linux-x86_64-server-release/images/jdk/bin/java -XX:CacheDataStore=app.cds -Xmx256m -Xms256m -XX:+UnlockExperimentalVMOptions -XX:+UseEpsilonGC  HelloStream.java
Warning: CacheDataStore cannot be used with default source launcher environment, disabling --add-modules=ALL-DEFAULT
hello, world
```

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/15/head:pull/15` \
`$ git checkout pull/15`

`$ git checkout pull/15` \
`$ git pull https://git.openjdk.org/leyden.git pull/15/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 15`

`$ git pr show -t 15`

</details>

<a href="https://git.openjdk.org/leyden/pull/15.diff">https://git.openjdk.org/leyden/pull/15.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/15#issuecomment-2333575097)

### DIFF
--- a/src/java.base/share/native/libjli/emessages.h
+++ b/src/java.base/share/native/libjli/emessages.h
@@ -77,6 +77,7 @@
 #define CFG_WARN3       "Warning: Missing VM type on line %d of `%s'"
 #define CFG_WARN4       "Warning: Missing server class VM on line %d of `%s'"
 #define CFG_WARN5       "Warning: Unknown VM type on line %d of `%s'"
+#define CFG_WARN6       "Warning: CacheDataStore cannot be used with default source launcher environment, disabling --add-modules=ALL-DEFAULT"
 
 #define CFG_ERROR1      "Error: Corrupt jvm.cfg file; cycle in alias list."
 #define CFG_ERROR2      "Error: Unable to resolve VM alias %s"
--- a/src/java.base/share/native/libjli/java.c
+++ b/src/java.base/share/native/libjli/java.c
@@ -1550,10 +1550,13 @@ ParseArguments(int *pargc, char ***pargv,
     }
 
     if (mode == LM_SOURCE) {
-        // Leyden: Supplying add-modules would break full module graph dumping,
-        // which will prevent CacheDataStore use. There is no other way to use
-        // CacheDataStore with source launcher at the moment. TODO: Figure this out.
-        if (!cacheDataStore) {
+        // Supplying add-modules would break full module graph dumping, which will
+        // prevent CacheDataStore use. There is no other way to use CacheDataStore
+        // with source launcher at the moment. Drop the add-modules, print warning
+        // message, and hope for the best. // TODO: Figure this out.
+        if (cacheDataStore) {
+          JLI_ReportErrorMessage(CFG_WARN6);
+        } else {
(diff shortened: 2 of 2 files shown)


vnkozlov:
We are trying  to avoid change Java launcher to check Leyden flags (especially -XX:) if we can do changes in VM.
See discussion in PR: https://github.com/openjdk/jdk/pull/20516

shipilev:
> So _not_ doing `--add-modules` when CDS is enabled is a convenience for performance testing, plus a breadcrumb for future work.

We can move this check to VM, but it is not convenient to figure our if we are running in `LM_SOURCE` mode or not on the VM side.

vnkozlov:
Lets wait @iklam comment on this.

iklam:
I am not sure if Leyden should be targeting running the launcher with a Java source file. There are many things that we *could* do to make some things run a little faster, but I think we should resist the temptation in this case.

shipilev:
I think what Leyden _targets_ is a different question.

After a day of playing with this, I think enabling source launcher does look very convenient for current performance testing, as it includes javac automatically, without any additional setup. I am not insisting on doing this workaround, it is just a handy development aid. If you all feel strongly against it, I can go the longer and less convenient route to measure CDS perf.

iklam:
But you will be adding javac timing to every one of your benchmarks. Is this really what you want?

If you want convenience, I think it's better to write a simple/generic script that compiles the Java source into a JAR file and then do it the usual way.

shipilev:
> But you will be adding javac timing to every one of your benchmarks. Is this really what you want?

Yes, it is exactly that: the easiest way for me to target javac is to add `.java` to my command line :)

shipilev:
No need for this anymore. We can just do javac tests directly.

PR #17, opened 2024-09-06
Prefer to take AOT method from the SC queue faster

shipilev:
Probably WIP. Humor me for a second: why could not / should not we take the AOT task for "compilation" (installation, really) right away like this? I suspect when SC queues are overwhelmed with N elements, we keep walking the entire SC queue for no particular reason?

This improves javac benchmark considerably (look at both wall time and user time):

```
# BEFORE
Benchmark 1: build/linux-x86_64-server-release/images/jdk/bin/java -XX:CacheDataStore=app.cds -Xmx256m -Xms256m -XX:+UnlockExperimentalVMOptions -XX:+UseEpsilonGC -cp JavacBenchApp.jar JavacBenchApp 50
  Time (mean ± σ):     348.3 ms ±   2.8 ms    [User: 949.0 ms, System: 101.4 ms]
  Range (min … max):   344.0 ms … 355.3 ms    100 runs

# AFTER
Benchmark 1: build/linux-x86_64-server-release/images/jdk/bin/java -XX:CacheDataStore=app.cds -Xmx256m -Xms256m -XX:+UnlockExperimentalVMOptions -XX:+UseEpsilonGC -cp JavacBenchApp.jar JavacBenchApp 50
  Time (mean ± σ):     312.2 ms ±   2.6 ms    [User: 577.8 ms, System: 95.4 ms]
  Range (min … max):   307.1 ms … 318.5 ms    100 runs
```

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/17/head:pull/17` \
`$ git checkout pull/17`

`$ git checkout pull/17` \
`$ git pull https://git.openjdk.org/leyden.git pull/17/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 17`

`$ git pr show -t 17`

</details>

<a href="https://git.openjdk.org/leyden/pull/17.diff">https://git.openjdk.org/leyden/pull/17.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/17#issuecomment-2334719677)

### DIFF
--- a/src/hotspot/share/compiler/compilationPolicy.cpp
+++ b/src/hotspot/share/compiler/compilationPolicy.cpp
@@ -907,6 +907,12 @@ CompileTask* CompilationPolicy::select_task(CompileQueue* compile_queue, JavaThr
       continue;
     }
     Method* method = task->method();
+    if (task->is_scc()) {
+      // AOT task, it should load fast, take it, and go.
+      max_task = task;
+      max_method = method;
+      break;
+    }
     methodHandle mh(THREAD, method);
     if (task->can_become_stale() && is_stale(t, TieredCompileTaskTimeout, mh) && !is_old(mh)) {
       if (PrintTieredEvents) {
--- a/src/hotspot/share/compiler/compilationPolicy.cpp
+++ b/src/hotspot/share/compiler/compilationPolicy.cpp
@@ -1188,10 +1188,7 @@ bool CompilationPolicy::compare_methods(Method* x, Method* y) {
 }
 
 bool CompilationPolicy::compare_tasks(CompileTask* x, CompileTask* y) {
-  if (x->is_scc() && !y->is_scc()) {
-    // x has cached code
-    return true;
-  }
+  assert(!x->is_scc() && !y->is_scc(), "SC tasks are not expected here");
   if (x->compile_reason() != y->compile_reason() && y->compile_reason() == CompileTask::Reason_MustBeCompiled) {
     return true;
   }
--- a/src/hotspot/share/compiler/compilationPolicy.cpp
+++ b/src/hotspot/share/compiler/compilationPolicy.cpp
@@ -906,13 +906,12 @@ CompileTask* CompilationPolicy::select_task(CompileQueue* compile_queue, JavaThr
       task = next_task;
       continue;
     }
-    Method* method = task->method();
     if (task->is_scc()) {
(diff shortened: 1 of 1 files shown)


vnkozlov:
Currently we do that in `compare_task`:
https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/compiler/compilationPolicy.cpp#L1185

I was very conservative to immediately selecting SCC task because we put these task on normal compilation queues:
https://github.com/veresov/jdk/commit/76157f6e77346ef0b2c47497231210f1fd84b446

Now we have separate queues for these tasks. I thought I changed code how task is selected for these queues (using FIFO).

src/hotspot/share/compiler/compilationPolicy.cpp
```diff
@@ -907,6 +907,12 @@ CompileTask* CompilationPolicy::select_task(CompileQueue* compile_queue, JavaThr
       continue;
     }
     Method* method = task->method();
+    if (task->is_scc()) {
+      // AOT task, it should load fast, take it, and go.
+      max_task = task;
+      max_method = method;
+      break;
```
vnkozlov, line 914: I think we can just return `task` here.

We don't cache `CompLevel_full_profile` Level3 C1 compiled code and as result we do nothing in the rest of this method for AOT code.

We put SCC tasks on separate queues which will have only these tasks.

shipilev: Right, that makes sense. Since we want to get the SC tasks as soon as possible, it makes sense to shortcut hard. I'll do it in new commit.

shipilev:
> Currently we do that in `compare_task`:
> https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/compiler/compilationPolicy.cpp#L1185

Aha! That is not efficient, since we are still walking the entire queue. I measured that task selection on 1K queue takes ~25us, which means for 2K elements we loiter in this code for 50ms, delaying the AOT loading. This explains why this patch has such an effect.

I dropped the SCC checks from `compare_task` and replaced them with asserts, since we are not expecting SCC tasks in that comparator anymore. No need to check SCC for normal compilation queues, which might delay compilation a bit.

vnkozlov (approved):
Good.

shipilev:
Thanks!

/integrate

PR #20, opened 2024-09-11
Load coops base shift from AOTRuntimeConstants in AOT code

adinn:
This PR modifies AOT compiled method code to load compressed oops base and shift constants via the AOTRuntimeConstants area rather than encode them as immediates. It also unlatches the currently forced setting of UseCompatibleCompressedOops, allowing the heap to be allocated wherever it will fit.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/20/head:pull/20` \
`$ git checkout pull/20`

`$ git checkout pull/20` \
`$ git pull https://git.openjdk.org/leyden.git pull/20/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 20`

`$ git pr show -t 20`

</details>

<a href="https://git.openjdk.org/leyden/pull/20.diff">https://git.openjdk.org/leyden/pull/20.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/20#issuecomment-2343781609)

### DIFF
--- a/src/hotspot/cpu/aarch64/aarch64.ad
+++ b/src/hotspot/cpu/aarch64/aarch64.ad
@@ -5172,7 +5172,7 @@ operand indOffX2P(iRegL reg, immLOffset off)
 
 operand indirectN(iRegN reg)
 %{
-  predicate(CompressedOops::shift() == 0);
+  predicate(CompressedOops::shift() == 0 && !SCCache::is_on_for_write());
   constraint(ALLOC_IN_RC(ptr_reg));
   match(DecodeN reg);
   op_cost(0);
@@ -5187,7 +5187,7 @@ operand indirectN(iRegN reg)
 
 operand indIndexScaledI2LN(iRegN reg, iRegI ireg, immIScale scale)
 %{
-  predicate(CompressedOops::shift() == 0 && size_fits_all_mem_uses(n->as_AddP(), n->in(AddPNode::Offset)->in(2)->get_int()));
+  predicate(CompressedOops::shift() == 0 && size_fits_all_mem_uses(n->as_AddP(), n->in(AddPNode::Offset)->in(2)->get_int()) && !SCCache::is_on_for_write());
   constraint(ALLOC_IN_RC(ptr_reg));
   match(AddP (DecodeN reg) (LShiftL (ConvI2L ireg) scale));
   op_cost(0);
@@ -5202,7 +5202,7 @@ operand indIndexScaledI2LN(iRegN reg, iRegI ireg, immIScale scale)
 
 operand indIndexScaledN(iRegN reg, iRegL lreg, immIScale scale)
 %{
-  predicate(CompressedOops::shift() == 0 && size_fits_all_mem_uses(n->as_AddP(), n->in(AddPNode::Offset)->in(2)->get_int()));
+  predicate(CompressedOops::shift() == 0 && size_fits_all_mem_uses(n->as_AddP(), n->in(AddPNode::Offset)->in(2)->get_int()) && !SCCache::is_on_for_write());
   constraint(ALLOC_IN_RC(ptr_reg));
   match(AddP (DecodeN reg) (LShiftL lreg scale));
   op_cost(0);
@@ -5217,7 +5217,7 @@ operand indIndexScaledN(iRegN reg, iRegL lreg, immIScale scale)
 
 operand indIndexI2LN(iRegN reg, iRegI ireg)
 %{
-  predicate(CompressedOops::shift() == 0);
+  predicate(CompressedOops::shift() == 0  && !SCCache::is_on_for_write());
   constraint(ALLOC_IN_RC(ptr_reg));
   match(AddP (DecodeN reg) (ConvI2L ireg));
   op_cost(0);
@@ -5232,7 +5232,7 @@ operand indIndexI2LN(iRegN reg, iRegI ireg)
 
 operand indIndexN(iRegN reg, iRegL lreg)
 %{
-  predicate(CompressedOops::shift() == 0);
+  predicate(CompressedOops::shift() == 0 && !SCCache::is_on_for_write());
   constraint(ALLOC_IN_RC(ptr_reg));
   match(AddP (DecodeN reg) lreg);
   op_cost(0);
@@ -5247,7 +5247,7 @@ operand indIndexN(iRegN reg, iRegL lreg)
 
 operand indOffIN(iRegN reg, immIOffset off)
 %{
-  predicate(CompressedOops::shift() == 0);
+  predicate(CompressedOops::shift() == 0 && !SCCache::is_on_for_write());
   constraint(ALLOC_IN_RC(ptr_reg));
   match(AddP (DecodeN reg) off);
   op_cost(0);
@@ -5262,7 +5262,7 @@ operand indOffIN(iRegN reg, immIOffset off)
 
 operand indOffLN(iRegN reg, immLOffset off)
 %{
-  predicate(CompressedOops::shift() == 0);
+  predicate(CompressedOops::shift() == 0 && !SCCache::is_on_for_write());
   constraint(ALLOC_IN_RC(ptr_reg));
   match(AddP (DecodeN reg) off);
   op_cost(0);
@@ -7790,7 +7790,7 @@ instruct convP2I(iRegINoSp dst, iRegP src) %{
 // in case of 32bit oops (heap < 4Gb).
 instruct convN2I(iRegINoSp dst, iRegN src)
 %{
-  predicate(CompressedOops::shift() == 0);
+  predicate(CompressedOops::shift() == 0 && !SCCache::is_on_for_write());
   match(Set dst (ConvL2I (CastP2X (DecodeN src))));
 
   ins_cost(INSN_COST);
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -5135,6 +5135,9 @@ void MacroAssembler::encode_heap_oop_for_aot(Register dst, Register src) {
   Register tmp = pick_different_tmp(dst, src);
   RegSet regs = RegSet::of(tmp);
 
+  // n.b. this relies on a cmp+csel sequence instead of a cbnz but
+  // only because the C2 encoding declares that it kills the flags
+  // register
   push(regs, sp);
   lea(tmp, ExternalAddress(AOTRuntimeConstants::coops_base_address()));
   ldr(tmp, tmp);
@@ -5166,15 +5169,19 @@ void MacroAssembler::decode_heap_oop_for_aot(Register dst, Register src) {
   Register tmp = pick_different_tmp(dst, src);
   RegSet regs = RegSet::of(tmp);
 
+  // n.b. this uses a branch instead of cmp+csel to avoid side-effects
+  // to the flags register.
   push(regs, sp);
+  mov(dst, src);
+  Label done;
+  cbz(dst, done);
   lea(tmp, ExternalAddress(AOTRuntimeConstants::coops_shift_address()));
   ldrw(tmp, tmp);
-  lslv(dst, src, tmp);
+  lslv(dst, dst, tmp);
   lea(tmp, ExternalAddress(AOTRuntimeConstants::coops_base_address()));
   ldr(tmp, tmp);
-  cmp(dst, zr);
-  csel(tmp, zr, tmp, EQ);
   add(dst, dst, tmp);
+  bind(done);
   pop(regs, sp);
 }
 
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -4970,6 +4970,10 @@ void MacroAssembler::encode_heap_oop(Register d, Register s) {
   verify_heapbase("MacroAssembler::encode_heap_oop: heap base corrupted?");
 #endif
   verify_oop_msg(s, "broken oop in encode_heap_oop");
+  if (SCCache::is_on_for_write()) {
+    encode_heap_oop_for_aot(d, s);
+    return;
+  }
   if (CompressedOops::base() == nullptr) {
     if (CompressedOops::shift() != 0) {
       assert (LogMinObjAlignmentInBytes == CompressedOops::shift(), "decode alg wrong");
@@ -5003,6 +5007,10 @@ void MacroAssembler::encode_heap_oop_not_null(Register r) {
   }
 #endif
   verify_oop_msg(r, "broken oop in encode_heap_oop_not_null");
+  if (SCCache::is_on_for_write()) {
+    encode_heap_oop_not_null_for_aot(r, r);
+    return;
+  }
   if (CompressedOops::base() != nullptr) {
     sub(r, r, rheapbase);
   }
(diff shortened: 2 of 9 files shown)


src/hotspot/share/code/SCCache.hpp
```diff
@@ -618,6 +618,9 @@ class AOTRuntimeConstants {
  friend class SCCache;
   uint _grain_shift;
   uint _card_shift;
+  address _coops_base;
```
vnkozlov, line 621: Move `address` first for better fields alignment

src/hotspot/share/code/SCCache.cpp
```diff
@@ -676,10 +676,13 @@ bool SCConfig::verify(const char* cache_path) const {
     log_warning(scc, init)("Disable Startup Code Cache: '%s' was created with RestrictContended = %s", cache_path, RestrictContended ? "false" : "true");
     return false;
   }
-  if (_compressedOopShift != (uint)CompressedOops::shift()) {
-    log_warning(scc, init)("Disable Startup Code Cache: '%s' was created with CompressedOops::shift() = %d vs current %d", cache_path, _compressedOopShift, CompressedOops::shift());
+#ifdef _LP64
+  if (UseCompatibleCompressedOops && (_compressedOopShift != (uint)CompressedOops::shift())) {
+    log_warning(scc, init)("Disable Startup Code Cache: '%s' was created with CompressedOops::shift() = %d vs current %d and UseCompatibleCompressedOops=%s", cache_path, _compressedOopShift, CompressedOops::shift(),
+                           UseCompatibleCompressedOops ? "true" : "false");
```
vnkozlov, line 682: Why print flags value when you check it in the condition - it will be `true`

src/hotspot/cpu/aarch64/aarch64.ad
```diff
@@ -5172,7 +5172,7 @@ operand indOffX2P(iRegL reg, immLOffset off)
 
 operand indirectN(iRegN reg)
 %{
-  predicate(CompressedOops::shift() == 0);
+  predicate(CompressedOops::shift() == 0 && !SCCache::is_on_for_write());
```
vnkozlov, line 5175: Does it imply that with SCC is on_write we will not have the shift == 0?

ashu-mehra: No, this is to prevent the match when compiling AOT code. We don't want to generate any AOT code which depends on shift == 0 as it may change during runtime.

vnkozlov: Okay.

vnkozlov:
Sorry, I am not sure about these changes any more vs enabling unconditionally `UseCompatibleCompressedOops`.
You will always use these loads, even for simplest "Unscaled" case. The performance will be much worse than using enforced constant shift and base.

Did you run any performance testing?

vnkozlov:
Thinking more.

On other hand, only AOT code will suffer. Recompiled code will be optimal which is good for peak performance.
And with UseCompatibleCompressedOops recompiled code will also suffer.

I need to see numbers how it affect performance of AOT code.  I hope CPU's prefetcher will keep these values nearby.

adinn:
@vnkozlov I pushed fixes for your feedback and then planned to get some performance numbers. However, when I rebuilt (on aarch64) and retested a simple javac HelloWorld I got a ClassCastEception from the production run:
```
java.lang.ClassCastException: class java.util.HashMap$Node cannot be cast to class java.util.HashMap$TreeNode (java.util.HashMap$Node and java.util.HashMap$TreeNode are in module java.base of loader 'bootstrap')
```

This is happening in the interpreter when executing HashMap::putVal in this code
```
            Node<K,V> e; K k;
            if (p.hash == hash &&
                ((k = p.key) == key || (key != null && key.equals(k))))
                e = p;
            else if (p instanceof TreeNode)
                e = ((TreeNode<K,V>)p).putTreeVal(this, tab, hash, key, value);
            else {
                for (int binCount = 0; ; ++binCount) {
```
The callout to the exception handler happens at bytecode 114
```
     104: aload         7
     106: instanceof    #173                // class java/util/HashMap$TreeNode
     109: ifeq          131
     112: aload         7
     114: checkcast     #173                // class java/util/HashMap$TreeNode
     117: aload_0
     118: aload         6
```
I looked at the oop in local 7 and it is indeed an instance of HashMap$Node, not HashMap$Treenode. We load both C1 and C2 code from the AOT code cache for this method. So, I supsect this may be a deopt issue. I'll try to catch the deopt callout and see if I can work out where it might be going wrong.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -5096,6 +5119,87 @@ void  MacroAssembler::decode_heap_oop_not_null(Register dst, Register src) {
...
 
+static Register pick_different_tmp(Register dst, Register src) {
+  Register tmp = r0;
+  if (tmp == src || tmp == dst) {
+    tmp = r1;
+    if (tmp == src || tmp == dst) {
+      tmp = r2;
+    }
+  }
+  return tmp;
```
theRealAph, line 5130:
```suggestion
  auto tmps = RegSet::of(r0, r1, r2) - RegSet::of(src, dest);
  return tmps.first();
```
Use the force, Luke Skywalker.

adinn: A-and ... this version actually works:

    auto tmps = RegSet::of(r0, r1, r2) - RegSet::of(src, dst);
    return *tmps.begin();

(`first()` is private :-)

adinn:
I tested any effects on performance by running the javac new workflow benchmark on my Linux/M2(aarch64) release build. Each timing is an average over 10 runs with runs for the different cases interleaved to amortize variation thanks to external factors. I ran the benchmark 3 times and found that the timings were not very consistent, leaving little room to pin anything down to the patch.

jvm1 = without coops
jvm2 = with coops patch

Run 1:
    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    261.08 ms
    [1_xon ] Premain JDK (CDS enabled)                     144.01 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       127.42 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  78.80 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    256.45 ms
    [2_xon ] Premain JDK (CDS enabled)                     145.14 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       125.20 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  80.53 ms
    ================================================================

Run 2:
    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    261.16 ms
    [1_xon ] Premain JDK (CDS enabled)                     149.71 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       134.25 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  84.16 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    262.24 ms
    [2_xon ] Premain JDK (CDS enabled)                     160.62 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       134.09 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  85.02 ms
    ================================================================

Run 3:
    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    266.63 ms
    [1_xon ] Premain JDK (CDS enabled)                     152.51 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       131.13 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  89.70 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    257.33 ms
    [2_xon ] Premain JDK (CDS enabled)                     147.52 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       128.13 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  86.08 ms
    ================================================================

This is running on bare metal with 64GB of RAM. So, the heap is allocated at 0x400000000 in boh jvm1 and jvm2 i.e. non-AOT code uses a zero base with shift 3. Only the 2_aot case can be affected by the patch but there is no noticeable difference wrt to 1_aot modulo the existing variation in results.

I reran with -Xms24M, allowing jvm2 to allocate the heap in low memory i.e. with zero for both shift and base (jvm1 still is still forced to allocate at an address > 4GB and use a shift). If anything this could only improve performance for all jvm2 cases relative to jvm1 (but less so for the 2_aot case).

Run 1:
    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    273.05 ms
    [1_xon ] Premain JDK (CDS enabled)                     148.72 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       131.19 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  80.10 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    264.14 ms
    [2_xon ] Premain JDK (CDS enabled)                     159.66 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       124.52 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  75.03 ms
    ================================================================

Run 2:
    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    267.48 ms
    [1_xon ] Premain JDK (CDS enabled)                     163.18 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       136.67 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  86.41 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    268.73 ms
    [2_xon ] Premain JDK (CDS enabled)                     157.70 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       136.69 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  90.39 ms
    ================================================================

Run 3:
    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    268.36 ms
    [1_xon ] Premain JDK (CDS enabled)                     161.01 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       134.43 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  91.79 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    260.27 ms
    [2_xon ] Premain JDK (CDS enabled)                     165.10 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       138.31 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  88.61 ms
    ================================================================

Once again there is no noticeable difference modulo the existing variation.

The results are not very convincing given the varying timings. However, as a tentative conclusion we can say:

1) The patch incurs no measurable loss in performance when the JVM is force to use same coops mode
2) The patch does not enable any measurable gain in performance by allowing the  JVM is force to use a more efficient coops mode.

So, I think we can drop this patch and stick with enforcing a compatible oops mode.

adinn:
Ah, ignore those results -- I had not configured the correct task set so it was mixing big and little CPUs. I'll repost corrected results.

adinn:
Ok, so here are the proper results, first when the heap is defaulted:

Run 1:

    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    235.10 ms
    [1_xon ] Premain JDK (CDS enabled)                     111.21 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       109.45 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  56.87 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    233.21 ms
    [2_xon ] Premain JDK (CDS enabled)                     111.37 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       107.96 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  59.00 ms
    ================================================================

Run 2:

    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    235.84 ms
    [1_xon ] Premain JDK (CDS enabled)                     112.28 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       108.45 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  57.76 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    236.09 ms
    [2_xon ] Premain JDK (CDS enabled)                     111.91 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       108.86 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  59.80 ms
    ================================================================


Run 3:


    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    231.32 ms
    [1_xon ] Premain JDK (CDS enabled)                     111.47 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       108.92 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  56.26 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    233.96 ms
    [2_xon ] Premain JDK (CDS enabled)                     114.21 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       107.92 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  60.01 ms
    ================================================================

Now the results when heap size is set to 24Mb:

Run 1:

    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    237.26 ms
    [1_xon ] Premain JDK (CDS enabled)                     114.10 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       108.43 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  56.43 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    238.70 ms
    [2_xon ] Premain JDK (CDS enabled)                     115.12 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       107.67 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  58.27 ms
    ================================================================

Run 2:

    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    237.81 ms
    [1_xon ] Premain JDK (CDS enabled)                     114.76 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       107.24 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  56.57 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    237.65 ms
    [2_xon ] Premain JDK (CDS enabled)                     115.34 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       107.66 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  59.39 ms
    ================================================================

Run 3:

    ============================== jvm1 ============================
    [1_xoff] Premain JDK (CDS disabled)                    238.08 ms
    [1_xon ] Premain JDK (CDS enabled)                     114.87 ms
    [1_td  ] Premain Prototype (CDS + Training Data)       107.35 ms
    [1_aot ] Premain Prototype (CDS + Training Data + AOT)  56.20 ms
    ============================== jvm2 ============================
    [2_xoff] Premain JDK (CDS disabled)                    237.19 ms
    [2_xon ] Premain JDK (CDS enabled)                     114.43 ms
    [2_td  ] Premain Prototype (CDS + Training Data)       107.40 ms
    [2_aot ] Premain Prototype (CDS + Training Data + AOT)  58.24 ms
    ================================================================

So, there is stil some variability in the timings which makes it hard to gauge the full effect. However, it looks as if the patch is adding at least 3% overhead to the run times for the aot case with the default heap and also with the smaller heap. The ability to vary the coops model with a smaller heap does not appear to win anything back either.

So, we definitely drop this patch.

vnkozlov:
Thank you @adinn for testing changes.

> So, we definitely drop this patch.

Agree.

ashu-mehra:
@adinn @vnkozlov We should also consider the fact that, as the things stand, forcing the JIT compiled methods to always use shift and fixed base could have an impact on the peak throughput. Do we have numbers how much is the peak throughput impacted? If it does then we need to decide what do we let go - startup or the peak throughput?

adinn:
Closing but keeping it available in case we find later that a better COOPs model can help improve peak performance when AOT code is replaced.

PR #21, opened 2024-09-11
8335358: [premain] Explore alternative ways to trigger the end of training run

macarte:
AOT training can be ended using either

- -XX: AOTEndTrainingOnMethodEntry =Hello.someMethod [same syntax as CompileOnly]
- -XX: AOTEndTrainingOnMethodEntry =Hello.someMethod,Hello.someOtherMethod,count=42
- jcmd \<pid> AOT.end_training

Also added programmatic hooks to System class:
- bool System.AOTIsTraining()
- void System.AOTEndTraining()

supports arm64 and x64

note: the AOTEndTrainingOnMethodEntry is ignored when not recording

JBS Issue: https://bugs.openjdk.org/browse/JDK-8335358

---------
### Progress

### Issue
 * [JDK-8335358](https://bugs.openjdk.org/browse/JDK-8335358): [premain] Explore alternative ways to trigger the end of training run (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/21/head:pull/21` \
`$ git checkout pull/21`

`$ git checkout pull/21` \
`$ git pull https://git.openjdk.org/leyden.git pull/21/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 21`

`$ git pr show -t 21`

</details>

<a href="https://git.openjdk.org/leyden/pull/21.diff">https://git.openjdk.org/leyden/pull/21.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/21#issuecomment-2344757092)
</details>


src/hotspot/share/services/diagnosticCommand.hpp
```diff
@@ -395,6 +395,24 @@ class ClassHierarchyDCmd : public DCmdWithParser {
...
+    static const char* name() { return "AOT.end_training"; }
+    static const char* description() {
+      return "End AOT training and create the cache.";
+    }
+    static const char* impact() {
+      return "High: [insert reason here]";
+    }
+    static const JavaPermission permission() {
+    JavaPermission p = {"java.lang.management.ManagementPermission",
+                        "monitor", nullptr};
```
macarte, line 410: Should this be 'monitor' or 'control' (or something else)?

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -1397,6 +1403,50 @@ methodHandle SharedRuntime::find_callee_method(TRAPS) {
...
+JRT_BLOCK_ENTRY(void, SharedRuntime::end_training_check_c1(JavaThread* current))
+{
+  JRT_BLOCK
+    SharedRuntime::end_training_check(CHECK);
+  JRT_BLOCK_END
+}
+JRT_END
+
+JRT_BLOCK_ENTRY(void, SharedRuntime::end_training_check_c2(JavaThread* current))
+{
```
macarte, line 1415: this and the _c1 method don't need to check the method flag to see if its a 'trigger', this is because we checked that on code generation which predicates the call to these methods

src/hotspot/share/interpreter/interpreterRuntime.cpp
```diff
@@ -1156,6 +1156,16 @@ void InterpreterRuntime::cds_resolve_invokedynamic(int raw_index,
   pool->cache()->set_dynamic_call(info, raw_index);
 }
 
+JRT_ENTRY(void, InterpreterRuntime::end_training_check(JavaThread* current)) {
+  LastFrameAccessor last_frame(current);
+  ResourceMark rm(current);
+  methodHandle m (current, last_frame.method());
+  if(m->is_end_training_trigger()) {
```
macarte, line 1163: unlike c1 and c2, we need to check the method flag here, as the generated code is used for many methods and so has to be included in all interpreter methods

src/hotspot/share/code/nmethod.hpp
```diff
@@ -514,10 +514,11 @@ class nmethod : public CodeBlob {
...
 
-  Method* method       () const { return _method; }
-  bool is_native_method() const { return _method != nullptr && _method->is_native(); }
-  bool is_java_method  () const { return _method != nullptr && !_method->is_native(); }
-  bool is_osr_method   () const { return _entry_bci != InvocationEntryBci; }
+  Method* method              () const { return _method; }
+  bool is_native_method       () const { return _method != nullptr && _method->is_native(); }
+  bool is_java_method         () const { return _method != nullptr && !_method->is_native(); }
+  bool is_osr_method          () const { return _entry_bci != InvocationEntryBci; }
+  bool is_end_training_trigger() const { return _method != nullptr && _method->is_end_training_trigger(); }
```
ashu-mehra, line 521: Is this used anywhere?

src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
```diff
@@ -1272,6 +1272,9 @@ address TemplateInterpreterGenerator::generate_native_entry(bool synchronized) {
   }
 #endif
 
+  // AOT training run support
+  __ end_training_check();
```
ashu-mehra, line 1276: Does it make sense to do these checks after `notify_method_entry()`? It could aid in debugging/testing that the training run indeed ended on this method entry.

ashu-mehra:
@macarte  Can we specify multiple methods each with its own value for `count` as triggers?

ashu-mehra (commented):
Minor suggestions

src/hotspot/share/code/nmethod.hpp
macarte: not anymore, will remove

src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
macarte: good call, will swap order of notify_method_entry and end_training_check

src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
```diff
@@ -1398,6 +1399,13 @@ void InterpreterMacroAssembler::_interp_verify_oop(Register reg, TosState state,
 
 void InterpreterMacroAssembler::verify_FPU(int stack_depth, TosState state) { ; }
 
+void InterpreterMacroAssembler::end_training_check() {
+  if (!FLAG_IS_DEFAULT(AOTEndTrainingOnMethodEntry) && CDSPreimage == nullptr) {
```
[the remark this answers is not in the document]
macarte: added a simple is_dumping_preimage_static_archive_with_triggers

macarte:
> @macarte Can we specify multiple methods each with its own value for `count` as triggers?

We decided to keep it simple for now; to support separate counts would require a map of methods to counts and using wildcards its possible to have many triggers e.g. java/lang/string.*.  Also while the compilers only add trigger_checks to the specified methods, due to the interpreter using a common method entry even a hash lookup would likely degrade the performance with so many calls

iklam:
I would suggest adding some sanity test cases. You can make a copy of test/hotspot/jtreg/runtime/cds/appcds/leyden/LeydenHello.java and rename the classes accordingly. Then, add the following to the `Tester` inner class:

```
        @Override
        public String[] vmArgs(RunMode runMode) {
            return new String[] {
                 "-XX: AOTEndTrainingOnMethodEntry=??????",
            };
        }
```

For jcmd, you can make a copy of test/hotspot/jtreg/runtime/cds/appcds/jcmd/JCmdTestStaticDump.java, strip most of it and leave only one copy of this:

```
        app  = createLingeredApp("-XX:CacheDataStore=foo.cds", "-cp", allJars);
        pid = app.getPid();
        test("foo.cds", pid, .....);
        app.stopApp();
```

You also need to modify JCmdTestDumpBase.java so it can dispatch the "AOT.end_training" command instead.

> note: the AOTEndTrainingOnMethodEntry is ignored when CDSPreImage is specified

I think this is OK. Filtering out the option from the command-line is non trivial; ignoring it is much simpler.

src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
```diff
@@ -1275,6 +1275,9 @@ address TemplateInterpreterGenerator::generate_native_entry(bool synchronized) {
   // jvmti support
   __ notify_method_entry();
 
+  // AOT training run support
+  __ end_training_check();
+
```
iklam, line 1280: I wonder how much this will slow down the interpreter, as we trap into the VM on every interpreter entry. Could you do something like

```
long start = System.currentTimeMillis();
for (int i = 0; i < 10000000; i++) {
    emptyMethod();
}
System.out.println(System.currentMillis() - start();

```
and run it with something like this? 

```
rm foo.cds
java -Xint  -XX:+UnlockDiagnosticVMOptions -XX:+CDSManualFinalImage -XX:CacheDataStore=foo.cds ...
```

src/hotspot/share/opto/parse1.cpp
```diff
@@ -1211,6 +1212,10 @@ void Parse::do_method_entry() {
     make_dtrace_method_entry(method());
   }
 
+  if (CDSConfig::is_dumping_preimage_static_archive_with_triggers() && method()->is_end_training_trigger()) {
```
iklam, line 1215: Do we come here also if `method()` is being inlined by C2?

src/hotspot/share/services/diagnosticCommand.hpp
iklam: I think "monitor" is appropriate (used also for dumping heap, etc). The API says "control" is to "control the runtime characteristics of the Java virtual machine", but we aren't really modifying the runtime characteristics.

macarte: Yes in v1 of this PR, ending the training run would cause the application to exit, hence our though on 'control', but as this is no longer the case, 'monitor' is more appropriate

src/hotspot/share/opto/parse1.cpp
macarte: It does, I tested inlining manually and proved that the method was inlined and still triggering, I'll see if I can add a test case so that if the inlining code path changes the test failure will direct us to add the appropriate code to the new code path

src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
macarte: The latest commit changes how we add triggers to the interpreter.  I added variants for the prologs that can have triggers, so only flagged methods with trap in the VM.  Also as the triggers are now only in flagged methods the VM code is faster as it know longer has to wrangle the method handle to test the flags.  While the traps remain even after the end of training has been triggered, the early out tests reduce the performance impact

ashu-mehra:
I think `is_end_training_trigger` needs to be cleared in `Method::remove_unshareable_flags()`. This flag should not be carried into the assembly phase if the Method is archived.

macarte:
> I think `is_end_training_trigger` needs to be cleared in `Method::remove_unshareable_flags()`. This flag should not be carried into the assembly phase if the Method is archived.

Thank you for the solution! I came across this problem while writing jtreg tests

rose00:
Suggestion:  Rename end_training_check in the interpreter to method_entry_upcall.  That way, the very wide but shallow changes to the assembly code will support a variety of features, not just this particular one.

Probably the same thing would work for the JITs as well.  So when C1 says

```
CDSConfig::is_dumping_preimage_static_archive_with_triggers() && callee->is_end_training_trigger()
```

Basically, the assembler and interpreter and JITs should not be thinking about CDS, just this newfangled "upcall".

it might as well say `callee->has_entry_upcall()` and the CDS-specific logic can be moved nearer to the method classes.  The name `is_end_training_trigger` can be retired or used in fewer places, when the only question at hand is whether to stick in an upcall.  Perhaps methodFlags continues to have `is_end_training_trigger` but `Method` exports `has_upcall` which delegates (internally) to `is_end_training_trigger`.

When you get to the C++ code for the upcall, just proceed as you do already, with a comment that the upcall has been installed if and only if the method was marked for training end.  Later on, that logic (again in C++) can be extended if we have other entry methods we wish to track.  Or not; maybe we'll never have a second use for this mechanism.  But even then, a more general looking name won't be harmful.  And there is some value in keeping the CDS-specific logic more localized, in a handful of functions that handle "upcalls" whatever those might be.

There are other lightweight versions of upcalls floating around too, low-level method entry notifications.  Some of them have been prototyped for Leyden (to trigger slow-burning recompilations from optimized code).  Factoring this feature into an upcall pattern might make it easier to merge the various kinds of upcalls into a common facility, eventually.

The name `_end_training_predicate` promises more than it is; shouldn't it be `_end_training_count_limit`?

I suggest passing the `CompileCommandEnum` into `parse_method_pattern`.  This will be a more user-facing instance of method patterns, so there will probably be special rules written for them.  Probably more restrictive than the present rules for method patterns in other compile commands.  For starters, we should choose between the dot-based and double-colon-based syntaxes.  I can't predict which way this conversation will go, but it will happen and we will be glad to be able to contextualize the parse to this particular command.

macarte:
Moving this PR to draft while we refactor the code as suggested

macarte:
> Suggestion: Rename end_training_check in the interpreter to method_entry_upcall. That way, the very wide but shallow changes to the assembly code will support a variety of features, not just this particular one.
> 
> Probably the same thing would work for the JITs as well. So when C1 says
> 
> ```
> CDSConfig::is_dumping_preimage_static_archive_with_triggers() && callee->is_end_training_trigger()
> ```
> 
> Basically, the assembler and interpreter and JITs should not be thinking about CDS, just this newfangled "upcall".
> 
> it might as well say `callee->has_entry_upcall()` and the CDS-specific logic can be moved nearer to the method classes. The name `is_end_training_trigger` can be retired or used in fewer places, when the only question at hand is whether to stick in an upcall. Perhaps methodFlags continues to have `is_end_training_trigger` but `Method` exports `has_upcall` which delegates (internally) to `is_end_training_trigger`.
> 
> When you get to the C++ code for the upcall, just proceed as you do already, with a comment that the upcall has been installed if and only if the method was marked for training end. Later on, that logic (again in C++) can be extended if we have other entry methods we wish to track. Or not; maybe we'll never have a second use for this mechanism. But even then, a more general looking name won't be harmful. And there is some value in keeping the CDS-specific logic more localized, in a handful of functions that handle "upcalls" whatever those might be.
> 
> There are other lightweight versions of upcalls floating around too, low-level method entry notifications. Some of them have been prototyped for Leyden (to trigger slow-burning recompilations from optimized code). Factoring this feature into an upcall pattern might make it easier to merge the various kinds of upcalls into a common facility, eventually.
> 
> The name `_end_training_predicate` promises more than it is; shouldn't it be `_end_training_count_limit`?
> 
> I suggest passing the `CompileCommandEnum` into `parse_method_pattern`. This will be a more user-facing instance of method patterns, so there will probably be special rules written for them. Probably more restrictive than the present rules for method patterns in other compile commands. For starters, we should choose between the dot-based and double-colon-based syntaxes. I can't predict which way this conversation will go, but it will happen and we will be glad to be able to contextualize the parse to this particular command.

Thanks again for the feedback, I've updated this (now draft) PR with these comments in mind, please take a look - will leave as draft PR until discussed

src/hotspot/share/cds/cdsConfig.cpp
```diff
@@ -671,6 +671,10 @@ bool CDSConfig::is_dumping_preimage_static_archive() {
   return _is_dumping_static_archive && CacheDataStore != nullptr && CDSPreimage == nullptr;
 }
 
+bool CDSConfig::is_dumping_preimage_static_archive_with_triggers() {
+  return _is_dumping_static_archive && CacheDataStore != nullptr && CDSPreimage == nullptr && !FLAG_IS_DEFAULT(AOTEndTrainingOnMethodEntry);
```
ashu-mehra, line 675: This can be replaced with `is_dumping_preimage_static_archive() && !FLAG_IS_DEFAULT(AOTEndTrainingOnMethodEntry)`

src/hotspot/share/cds/cdsEndTrainingUpcall.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#ifndef SHARE_CDS_CDSENDTRAININGUPCALL_HPP
+#define SHARE_CDS_CDSENDTRAININGUPCALL_HPP
+
+#include "memory/allStatic.hpp"
+#include "utilities/globalDefinitions.hpp"
+
+
+#include "code/codeBlob.hpp"
+#include "code/vmreg.hpp"
```
ashu-mehra, line 33: umm...do we need all these includes here? I think some may be redundant.
Same applies to runtimeUpcallNop.hpp.

src/hotspot/share/runtime/runtimeUpcalls.hpp
```diff
@@ -0,0 +1,121 @@
...
+    _address = CAST_FROM_FN_PTR(address, upcall);
+  }
+
+private:
+  friend class RuntimeUpcalls;
+  void set_index(const int index) { _index = index; }
+  int get_index() const { assert(_index >= 0, "invalid index"); return _index; }
+
+public:
+  static RuntimeUpcallInfo* Create(const char* upcallName, const RuntimeUpcall upcall, const RuntimeUpcallMethodFilterCallback methodFilter) {
```
ashu-mehra, line 65: I think the convention is to always use lower case for methods, so probably s/Create/create

src/hotspot/share/runtime/runtimeUpcalls.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+  // This redirection occurs when there are more than one upcalls setup.  Currently each method is marked
+  // to indicate either none, entry and/or exit upcalls (two bits total); then we have to iterate over
+  // all upcalls and test the method details to determine which upcalls to call.  This is not optimal.
+  // One possible optimization is to use more bits to support more upcalls.  The method flags currently use 18
+  // out of 32 bits, so there are still 14 bits available for use.  We could set a limit of say 4-8 entry/exit
+  // upcalls combined, leaving 10-6 bits for other uses.  This still requires a redirect here to determine
+  // which upcalls to call, but it would be more efficient than the current implementation as we'd avoid the
+  // method matching and simply map bits to indexes.
+
```
ashu-mehra, line 96: Another way to speed it up is to store a hashset of Method pointers in each RuntimeUpcallInfo when `RuntimeUpcalls::mark_for_upcalls` is called for the method. Then the method matching can be replaced with a  lookup in the hashset.

src/hotspot/share/runtime/runtimeUpcallNop.cpp
```diff
@@ -0,0 +1,51 @@
...
+        RuntimeUpcallNop::nop_method,
+        RuntimeUpcallNop::filter_method_callback)) {
+    return true;
+  }
+  return false;
+}
+
+bool RuntimeUpcallNop::filter_method_callback(MethodDetails& methodDetails)
+{
+  return false;
```
ashu-mehra, line 45: Shouldn't this be returning true to enable it for all methods as per the documentation for AddRuntimeUpcallsNOP in globals.hpp?

src/hotspot/share/runtime/runtimeUpcalls.hpp
```diff
@@ -0,0 +1,121 @@
...
+#include "runtime/methodDetails.hpp"
+#include "utilities/macros.hpp"
+
+enum RuntimeUpcallType{
+  onMethodEntry = 0,
+  onMethodExit,
+  numTypes
+};
+
+typedef void (*RuntimeUpcall)(JavaThread* current);
```
ashu-mehra, line 39: Not needed for the AOTEndTrainingOnMethodEntry option, but I think passing the method as the argument for which upcall is being done could be beneficial.

macarte:
> @macarte The RuntimeUpcalls has support for method exit upcalls, but the changes in interpreter and compilers are missing. I guess this is intentional as there is no use-case for the method exit upcalls as of now.

That's correct, there's no use-case but I added the option so that a future implementation doesn't require a change to all existing upcalls.  I also added an assert that will trigger if an attempt to register an onMethodExit upcall is made.

src/hotspot/share/runtime/runtimeUpcallNop.cpp
macarte: Thanks for pointing that out; I've updated the AddRuntimeUpcallsNOP to a string that is of the format "=[onMethodEntry|onMethodExit]:[all|none]" allowing us to test (a) mulitple upcalls ,(b) test onMethodExit (that it currently asserts), (c) have the NOP upcall do nothing, but either install an upcall on no methods or all methods

macarte: So now the filter will use the value from the string 'none' (false) or 'all' (true) for the filter result

src/hotspot/share/runtime/runtimeUpcalls.hpp
macarte: I can see where that might be useful, but it's complicated as we have a single call back but the 'method' comes back from the interpreter as a ciMethod instance and from c1/c2 as a Method instance, so we'd need to either have a level of indirection so we could wrap the method (like I did with MethodDetails) or have multiple callbacks

macarte:
Thanks @ashu-mehra for the review, I've addressed each of your comments (some with a code change)

src/hotspot/share/cds/cdsEndTrainingUpcall.cpp
```diff
@@ -0,0 +1,146 @@
...
+      MetaspaceShared::preload_and_dump(current);
+      assert(!current->has_pending_exception(), "Unexpected exception");
+      return true;
+    }
+  }
+  return false;
+}
+
+bool CDSEndTrainingUpcall::filter_method_callback(MethodDetails& methodDetails)
+{
```
iklam, line 86: The HotSpot convention would be to use `method_details` instead of `methodDetails` for the names of fields, parameters and local variables.

See https://github.com/openjdk/jdk/blob/master/doc/hotspot-style.md#naming

macarte: Thanks, I've corrected all camelCase to lowercase_lowercase format

iwanowww:
Nice work, Mat! 

Speaking of proposed API, there's a lot of complexity around `-XX:AOTEndTrainingOnMethodEntry=`. What's the primary use case for it? Does it add much for uncooperative case (no application changes) compared to `jcmd`?

For cooperative case, a dedicated method in JDK an application can call seems more appropriate.

PR #25, opened 2024-10-23
Save/load i2c2i adapters

ashu-mehra:
This is an attempt to save and load i2c2i adapters along with the adapter handler table.
There are mainly two parts to this change:
1. Storing of adapter code in the SCCache or AOT code cache.
2. Storing of adapter handler table in the AOT cache.

Adapter handler table is a map from AdapterFingerPrint to AdapterHnadlerEntry. To store them in AOT cache, AdapterFingerPrint and AdapterHandlerEntry are updated to MetaspaceObj. Both these entities are discovered and added to the cache while processing the Method. When storing the adapter handler table, only the entries that have already been archived are considered. This allows pruning of AdapterHnadlerEntry that may be only reachable through a Method that is not eligible to be archived. 

An AdapterHandlerEntry has pointer to the adapter code. Because the AdapterHandlerEntry and the adapter code are stored in separate archives, this link between the AdapterHandlerEntry and the adapter code needs to be removed (see AdapterHandlerEntry::remove_unshareable_info()).
During the production run, as the methods in the AOT cache are adopted, the AdapterHandlerEntry is linked back to the adapter code (see AdapterHandlerEntry::restore_unshareable_info).

All this code is guarded by -XX:[+-]ArchiveAdapters option which defaults to false, but is set to true in CDSConfig during the assembly phase.

Other changes worth mentioning:
1. Changes to the SCCache infrastructure to make it possible to store and load adapter code. (Thanks to @adinn)
2. Updating AdapterFingerPrint hashing algorithm to avoid collisions. If there is any collision, then it will prevent finding the adapter code in the SCCache. (Again courtesy of @adinn)

Thanks to @adinn for providing many of these changes.

Performance:
-Xlog:init shows time taken for linking of Methods and making adapters. An example output is:
```
ClassLoader:
  clinit:                             150us / 4612 events
  link methods:                     28980us / 176893 events
  method adapters:                  15378us / 697 events
```
Save/load of adapters seem to have improved these stats.

| Quarkus | -ArchiveAdapters | +ArchiveAdapters |
|---|---|---|
| link methods | 12214us / 58913 events | 2700us / 58913 events |
| method adapters | 7793us / 607 events | 4402us / 38 events |

| Spring-petclinic | -ArchiveAdapters | +ArchiveAdapters |
|---|---|---|
| link methods | 28980us / 176893 events | 7485us / 176893 events |
| method adapters | 15378us / 697 events | 7050us / 13 events |

However, testing with Quarkus app, I don't see any noticeable improvement in the startup time.

---------
### Progress

### Reviewers

### Contributors
 * Andrew Dinn `<adinn@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/25/head:pull/25` \
`$ git checkout pull/25`

`$ git checkout pull/25` \
`$ git pull https://git.openjdk.org/leyden.git pull/25/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 25`

`$ git pr show -t 25`

</details>

<a href="https://git.openjdk.org/leyden/pull/25.diff">https://git.openjdk.org/leyden/pull/25.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/25#issuecomment-2434066262)
</details>


src/hotspot/share/code/SCCache.cpp
```diff
@@ -580,10 +589,28 @@ SCCache::SCCache(const char* cache_path, int fd, uint load_size) {
   _table = new SCAddressTable();
 }
 
-void SCCache::init_table() {
+void SCCache::init_extrs_table() {
+  SCCache* cache = SCCache::cache();
+  if (cache != nullptr && cache->_table != nullptr) {
+    cache->_table->init_extrs();
```
vnkozlov, line 595: This is repetitive checks pattern. I suggest to replace it with check of call to:
```
  SCAddressTable* SCCache::addr_table() const {
    return is_on() && (cache()->_table != nullptr) ? cache()->_table : nullptr;
  }
```

src/hotspot/share/code/SCCache.cpp
```diff
@@ -1131,6 +1164,20 @@ bool SCCache::finish_write() {
...
           preload_entries[preload_entries_cnt++] = entries_count;
+        } else if (entries_address[i].kind() == SCCEntry::Adapter) {
+          adapters_count++;
+        } else if (entries_address[i].kind() == SCCEntry::Stub) {
+          stubs_count++;
+        } else if (entries_address[i].kind() == SCCEntry::Blob) {
+          total_blobs_count++;
+          if (entries_address[i].comp_level() == CompLevel_none) {
+            shared_blobs_count++;
+          } else if (entries_address[i].comp_level() == CompLevel_simple) {
```
vnkozlov, line 1175: we also caching C1 code with simple profiling: `CompLevel_limited_profile`

src/hotspot/share/code/SCCache.cpp
```diff
@@ -2289,6 +2481,11 @@ bool SCCache::store_exception_blob(CodeBuffer* buffer, int pc_offset) {
     buffer->decode();
   }
 #endif
+  // we need to take a lock to stop C1 and C2 compiler threads racing to
+  // write blobs in parallel with each other or with later nmethods
+  // TODO - maybe move this up to selected callers so we only lock
+  // when saving a c1 or opto blob
+  MutexLocker ml(Compile_lock);
```
vnkozlov, line 2488: Only one compiler thread should generate compiler's runtime blobs.
See `AbstractCompiler::should_perform_init()`.

vnkozlov: Unless you changed when you call this generation.

src/hotspot/share/oops/method.cpp
```diff
@@ -1256,17 +1266,21 @@ void Method::link_method(const methodHandle& h_method, TRAPS) {
     ClassLoader::perf_ik_link_methods_count()->inc();
   }
 
+  // TODO: how to identify code cache full situation now that the adapter() can be
+  // non-null if AOT cache is in use
+#if 0
```
vnkozlov, line 1271: Can you check next (opposite to check in the following assert)?:
```
 if (adapter() != nullptr && !adapter()->is_linked()) {
```
The assumption is that we have enough CodeCache when we loading adapters from APT cache. Otherwise we should bailout (did you test such case?).

Is `is_linked()` is specific for adapters from AOT cache?

vnkozlov (commented):
Very nice changes.  Few comments.

src/hotspot/share/oops/method.cpp
```diff
@@ -423,6 +427,10 @@ void Method::remove_unshareable_info() {
 
 void Method::restore_unshareable_info(TRAPS) {
   assert(is_method() && is_valid_method(this), "ensure C++ vtable is restored");
+  if (_adapter != nullptr) {
+    _adapter->restore_unshareable_info(CHECK);
+    _from_compiled_entry = _adapter->get_c2i_entry();
```
iklam, line 432: There's similar code in `Method::restore_archived_method_handle_intrinsic()`. I would suggest combining them into a helper function to avoid code duplication.

src/hotspot/share/code/SCCache.cpp
ashu-mehra: Done

src/hotspot/share/code/SCCache.cpp
ashu-mehra: Done

src/hotspot/share/oops/method.cpp
ashu-mehra: Done

src/hotspot/share/code/SCCache.cpp
ashu-mehra: Added check for `CompLevel_limited_profile`

src/hotspot/share/oops/method.cpp
ashu-mehra: `is_linked()` is being set for every AdapterHandlerEntry when its code is either generated or loaded from AOT cache.

Regarding the original block of code that this check pertains to:

```
  // If the code cache is full, we may reenter this function for the
  // leftover methods that weren't linked.
  if (adapter() != nullptr) {
    return;
  }
```

The comment seem to indicate that we may reenter this function for a Method* for which adapter code has already been generated. However I am not able to trace the code path that may result in re-entering this function. Can you please explain under what conditions is this possible? @vnkozlov

vnkozlov: The comment was added for [JDK-7033141](https://bugs.openjdk.org/browse/JDK-7033141)
https://hg.openjdk.org/jdk9/jdk9/hotspot/rev/d3b9f2be46ab

I think the comment is incorrect. It should talk about PermGen space based on bug's evaluation:

"If the VM runs out of permgen space while allocating the constant pool cache, it tries to reverify the bytecodes in the methods for the class. But the bytecodes have been rewritten. I'm working on a fix that un-rewrites the bytecodes so that the VM can try again to link this class. I am debugging this now - actually I'm debugging my code that forces the error condition (for testing) since this but only reproduces for a specific error condition.
It's not very unlikely for an application to run out of permgen (or code cache as in bug 6947901) so it is probably worth fixing for jdk 7. The fix is relatively low risk once it's debugged."

[JDK-6947901](https://bugs.openjdk.org/browse/JDK-6947901) shows failure with -Xint too. 

But I imaging that full CodeCache may also cause failure to create adapters which will cause "un-rewrites" bytecode.

We don't have PermGen anymore. The only issue is space in CodeCache for adapters. Which you can check before loading adapters since you know size of adapters code in AOT cache.

I don't think we currently check that CodeCache size is the same during product run as during AOT Assembly phase. Adapters are allocated in `NonNMethod` section.

ashu-mehra: @vnkozlov I added a change to fix this by checking if adapter is shared or not. If it is not shared and is not null, we return, else we continue. This should restore the behavior of returning early if link_method() gets called again due to code cache full.

vnkozlov: Good!

ashu-mehra:
@vnkozlov @iklam is this good to go now? Any more comments before I merge it?

vnkozlov:
Is this up to date with current `premain` branch code?

ashu-mehra:
@vnkozlov yes, I have merged premain into this PR"s branch. premain is now ahead of this by just two commits which were done yesterday.

vnkozlov:
@iklam can you run this patch through our internal testing?

iklam:
> @iklam can you run this patch through our internal testing?

OK I will do it.

iklam:
> > @iklam can you run this patch through our internal testing?
> 
> OK I will do it.

I am seeing new failures on aarch64 only. x64 seems fine:

```
runtime/cds/appcds/applications/JavacBench.java#leyden			macosx-aarch64-debug
runtime/cds/appcds/applications/MicronautFirstApp.java#leyden		macosx-aarch64-debug
runtime/cds/appcds/applications/MicronautFirstApp.java#leyden		linux-aarch64-open
runtime/cds/appcds/applications/MicronautFirstApp.java#leyden		linux-aarch64-debug
runtime/cds/appcds/applications/QuarkusGettingStarted.java#leyden       linux-aarch64-debug
```

Here's the hs_err. This happens in the final production run. 

```
#  SIGSEGV (0xb) at pc=0x0000ffff89976650, pid=3238003, tid=3238006
#
# JRE version: Java(TM) SE Runtime Environment (24.0) (fastdebug build 24-internal-2025-01-07-1758562.ioi.lam.le4)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (fastdebug 24-internal-2025-01-07-1758562.ioi.lam.le4, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
# Problematic frame:
# v  ~AdapterBlob 0x0000ffff83d1b0d8

---------------  S U M M A R Y ------------

Command Line: -XX:MaxRAMPercentage=6.25 -Dtest.boot.jdk=/opt/mach5/mesos/work_dir/jib-master/install/jdk/23/37/bundles/linux-aarch64/jdk-23_linux-aarch64_bin.tar.gz/jdk-23 -Djava.io.tmpdir=/opt/mach5/mesos/work_dir/slaves/b733f181-520a-4536-86fc-7df55263c942-S3131/frameworks/1735e8a2-a1db-478c-8104-60c8b0af87dd-0196/executors/4ab6604d-6790-4797-a358-fcfe3a5e2cfe/runs/b53804f1-a345-47c7-9bed-24313dccd140/testoutput/test-support/jtreg_open_test_hotspot_jtreg_hotspot_tier2_runtime/tmp -DautoQuit=true -Dmicronaut.server.port=0 -XX:+IgnoreUnrecognizedVMOptions -XX:-VerifyDependencies -XX:+UnlockDiagnosticVMOptions -XX:VerifyArchivedFields=2 -Xlog:cds:file=MicronautFirstApp.production.log::filesize=0 -XX:CacheDataStore=MicronautFirstApp.cds -Xlog:scc=error example.micronaut.Application
```

ashu-mehra:
@iklam I have tried reproducing these failures on an linux-aarch64 (fedora 40) system but the tests always pass. I have run them multiple times using fastdebug and release build but didn't get any failure.
Are these failure reproducible every time in your testing, or are they intermittent? Which Linux distro were this tests run on? Also, can you share the backtrace for the crash.

iklam:
> @iklam I have tried reproducing these failures on an linux-aarch64 (fedora 40) system but the tests always pass. I have run them multiple times using fastdebug and release build but didn't get any failure.
> 
> Are these failure reproducible every time in your testing, or are they intermittent? Which Linux distro were this tests run on? Also, can you share the backtrace for the crash.

I got those crashes from our CI pipeline. Let me try to run the tests manually on linux-aarch hosts and see if I can reproduce the problems.

adinn:
Interesting also that the same test takes around 2 minutes on a Mac M2 with or without Ashu's patch. So, whatever the patch is upsetting appears to be specific to x86 . . .

adinn:
Of course, the difference on aarch64 is that c1/aarch64 deopts when a patch is needed. So, there is no updating of live method code other than to poison the method entry and cause a SIGILL.

Also, I switched on TracePatching on x86_64 and the time came right back down again. Since that involves taking the tty lock each time a patch is attempted it suggests that this is to do with threads tripping over each other when they race to perform the patch, or maybe failing to correctly synchronize data or code changes?

adinn:
> Without the patch it finishes in just over 2 minutes. I am trying to debug this now.

Actually I compared against mainline jdk rather than premain. Then I saw the same failure in an unrelated Leyden patch (JDK-8348957). So, I think this may be an issue in the latest premain. I'm comparing against that now.

ashu-mehra:
@iklam @vnkozlov there have been some changes since last review to fix a number of bugs in this PR. So I am requesting another review.

vnkozlov (approved):
Looks fine to me.

iklam (approved):
LGTM

shipilev:
Pull from current `premain` to get the compile queue bugfix, that should turn GHAs green.

PR #26, opened 2024-12-09
8345836: stable annotation documentation is incomplete

rose00:
The javadoc for jdk.internal.vm.annotation.Stable is incomplete.

The existing documentation gives an over-simple user model,
and does not explain how it is implemented.
Proposed new documentation will detail how the annotation
is implemented, and how it may be used correctly.

The improved documentation will makes it easier for JDK programmers
to use the annotation more aggressively, and more confidently.

This is a first cut.  Please comment…

---------
### Progress

### Warning
&nbsp;⚠️ Found leading lowercase letter in issue title for `8345836: stable annotation documentation is incomplete`

### Issue
 * [JDK-8345836](https://bugs.openjdk.org/browse/JDK-8345836): stable annotation documentation is incomplete (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/26/head:pull/26` \
`$ git checkout pull/26`

`$ git checkout pull/26` \
`$ git pull https://git.openjdk.org/leyden.git pull/26/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 26`

`$ git pr show -t 26`

</details>

<a href="https://git.openjdk.org/leyden/pull/26.diff">https://git.openjdk.org/leyden/pull/26.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/26#issuecomment-2529868394)
</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
+++ b/src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
@@ -30,11 +30,13 @@ import java.lang.annotation.*;
 /**
  * A field may be annotated as "stable" to indicate that it is a
  * <em>stable variable</em>, expected to change its value just once.
- * While the field contains its initial default (null or zero) value,
- * the VM treats it as an ordinary mutable variable.  When the
- * first value is stored into the field (assuming it is not a
- * duplicate of the the field's default initial value), the VM may
- * assume that no more significant changes will occur.  This in
+ * All Java fields are initialized by the VM with a default value
+ * (null or zero).  A properly used stable field will be set
+ * just once to a non-default value, and keep that value forever.
+ * While the field contains its default (null or zero) value,
+ * the VM treats it as an ordinary mutable variable.  When a
+ * non-default value is stored into the field, the VM is permitted
+ * to assume that no more significant changes will occur.  This in
  * turn enables the VM to optimize uses of the stable variable, treating
  * them as constant values.  This behavior is a useful building block
  * for lazy evaluation or memoization of results.  In rare and subtle
@@ -65,11 +67,11 @@ import java.lang.annotation.*;
  * the VM discovers a stable variable holding a null or primitive zero
  * value.  Does the user intend the VM to constant fold that
  * (uninteresting) value?  Or is the user waiting until later to
- * assign another value to the variable?  The VM does not
- * systematically record stores of a default null (or primitive zero),
- * so there is no way for the VM to decide if a default field value is
- * an undisturbed initial default value, or has been overwritten with
- * an intentionally stored null (or primitive zero).  This is why the
+ * assign a permanent value to the variable?  The VM does not
+ * systematically record stores of a null (resp., zero) to a stable variable,
+ * so there is no way for the VM to decide if a field's current value is
+ * its undisturbed initial value, or has been overwritten with
+ * an intentionally stored null (resp., zero).  This is why the
  * programmer should store non-default values into stable variables,
  * if the consequent optimization is desired.
  * <p>
@@ -79,6 +81,15 @@ import java.lang.annotation.*;
  * stored and what races are possible, safe publication may require
  * special handling with a {@code VarHandle} atomic method.
  * (See below.)
+ * <p>
+ * If an application requires constant folding of a stable variable
+ * whose permanent value may be the default value (null or zero),
+ * the variable can be refactored to add an extra indirection.
+ * This would represent the default value in a non-null "box",
+ * such as {@code Integer.valueOf(0)} or a lambda like
+ * {@code ()->null}.  Such a refactoring should always be possible,
+ * since stable variables should (obviously) never be part of public
+ * APIs.
  *
  * <h2><a id="arrays"></a>Stable Arrays</h2>
  *
@@ -130,7 +141,7 @@ import java.lang.annotation.*;
  * <h2><a id="examples"></a>Examples of Stable Variables</h2>
  *
  * In the following example, the only constant-foldable string stored
- * in a stable value is the string {@code "S"}.  All subarrays are
+ * in any stable variable is the string {@code "S"}.  All subarrays are
  * constant.
  *
  * <pre>{@code
@@ -169,8 +180,9 @@ import java.lang.annotation.*;
  * }</pre>
  *
  * <p>
- * As a very simple example, a stable boolean variable (i.e., stable
- * field or stable boolean array element) might be constant-folded,
+ * Note that a stable boolean variable (i.e., a stable
+ * field like {@code INITIALIZED}, or a stable boolean
+ * array element above) can be constant-folded,
  * but only after it is set to {@code true}.  Even this simple
  * optimization is sometimes useful for responding to a permanent
  * one-shot state change, in such a way that the compiler can remove
@@ -179,16 +191,6 @@ import java.lang.annotation.*;
  * its default (i.e., {@code false}), since compiled code might have
  * captured the {@code true} value as a constant, and as long as that
  * compiled code is in use, the reset value will go undetected.
- * <p>
- * Stability of a variable and the constant folding of its default
- * value (null or zero) are mutually exclusive, as explained above.
- * So {@code final} field might be both null and also a constant, but
- * not a stable field.  If an application requires constant folding of
- * the default value, it can add an extra indirection to "box" the
- * default value, such as {@code Integer.valueOf(0)} or a lambda like
- * {@code ()->null}.  Such a refactoring should always be possible,
- * since stable variables should (obviously) never be part of public
- * APIs.
  *
  * <h2><a id="final"></a>Final Variables, Stable Variables, and Memory Effects</h2>
  *
--- a/src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
+++ b/src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
@@ -39,20 +39,26 @@ import java.lang.annotation.*;
  * them as constant values.  This behavior is a useful building block
  * for lazy evaluation or memoization of results.  In rare and subtle
  * use cases, stable variables may also assume multiple values over
- * time, with effects to be described below.
+ * time, with effects as described below.
  * <p>
- * (Warning: the {@code @Stable} annotation is intended for use in the
+ * <em>(Warning: the {@code @Stable} annotation is intended for use in the
  * JDK implemention, and with the HotSpot VM, to support optimization
  * of classes and algorithms defined by the JDK.  It is unavailable
- * outside the JDK.)
- * <p>
- * For example, declaring two stable fields of type {@code int} and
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
```diff
@@ -28,56 +28,191 @@
...
+ * (uninteresting) value?  Or is the user waiting until later to
+ * assign another value to the variable?  The VM does not
+ * systematically record stores of a default null (or primitive zero),
+ * so there is no way for the VM to decide if a default field value is
+ * an undisturbed initial default value, or has been overwritten with
+ * an intentionally stored null (or primitive zero).  This is why the
+ * programmer should store non-default values into stable variables,
+ * if the consequent optimization is desired.
+ * <p>
+ * As a special case, if a stable field is declared as an array type
```
minborg, line 63: If `array` is a stable array field, is `array.length` treated as an independent stable variable?

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
```diff
@@ -28,56 +28,191 @@
...
  * <p>
- * It is (currently) undefined what happens if a field annotated as stable
- * is given a third value (by explicitly updating a stable field, a component of
- * a stable array, or a final stable field via reflection or other means).
- * Since the HotSpot VM promotes a non-null component value to constant, it may
- * be that the Java memory model would appear to be broken, if such a constant
- * (the second value of the field) is used as the value of the field even after
- * the field value has changed (to a third value).
+ * After constant folding, the compiler can make use of may aspects of
+ * the object: Its dynamic type, its length (if it is an array), and
```
minborg, line 165: Ahh. Here is the answer to the `array.length` question. Good.

minborg:
It might be worth mentioning that primitive types could be held in a wrapper class (rather than as a primitive value) if the default values are important to capture and constant-fold.

Another thing that we could talk about when it comes to stable variables initialized outside `<init>` and `<clinit>` in a multi-threaded scenario is that code can use CAS operations when initializing the stable variable to uphold the update-at-most-once invariant and that readers can use acquire/release or volatile semantics when reading the stable variable. This ensures consistency across threads while still allowing VM optimizations.

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
```diff
@@ -28,56 +28,191 @@
...
  * VM.
  * <p>
- * It is (currently) undefined what happens if a field annotated as stable
- * is given a third value (by explicitly updating a stable field, a component of
- * a stable array, or a final stable field via reflection or other means).
- * Since the HotSpot VM promotes a non-null component value to constant, it may
- * be that the Java memory model would appear to be broken, if such a constant
- * (the second value of the field) is used as the value of the field even after
- * the field value has changed (to a third value).
+ * After constant folding, the compiler can make use of may aspects of
```
mcimadamore, line 164:
```suggestion
 * After constant folding, the compiler can make use of many aspects of
```

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
```diff
@@ -28,56 +28,191 @@
...
+ * Note also each compiliation task makes its own decisions about
+ * whether to observe stable variable values, and how aggressively to
+ * constant-fold them.  And a method that uses a stable variable might
+ * be inlined by many different compilation tasks.  The net result of
+ * all this is that, if stable variables are multiply assigned, the
+ * program execution may observe any "historical" value (if it was
+ * captured by some particular compilation task), as well as a "most
+ * recent" value observed by the interpreter or less-optimized code.
+ * <p>
+ * For all these reasons, a user who bends the rules for a stable
```
mcimadamore, line 196: I've cases in the JDK where a stable fields could, under race, be assigned to objects that are semantically equivalent, but whose identity is distinct. Would you say this is a good use of `@Stable` ? In FFM we took extra precaution and deduplicated the identity by using a concurrent hash map -- but is this required?

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
```diff
@@ -28,56 +28,191 @@
...
+ * implemention, and with the HotSpot VM, to support optimization of
+ * classes and algorithms defined by the JDK.  Any class which uses
+ * this annotation is responsible for constraining assignments in such
+ * a way as not to violate API contracts of the class.  Such
+ * constraints can be arranged using explicit atomic access
+ * (sychronization, CAS, etc.), or by concealing the effects of
+ * multiple assignments in some API-dependent way, or by providing
+ * some other internal proof of correctness (accounting for any
+ * possible racing API access), or by some appropriate disclaimer in
+ * the API about undefined behavior.
```
RogerRiggs, line 215: This caution/warning might be better at the top of the file.

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
liach: I think so; we have benign races for non-stable fields like `Class.reflectionFactory`. However, sometimes sharing the same object avoids redundant cache computation. It is ultimately a tradeoff between synchronization and lock overheads and the applicability of cache.

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
```diff
@@ -28,56 +28,191 @@
...
- * It is (currently) undefined what happens if a field annotated as stable
- * is given a third value (by explicitly updating a stable field, a component of
- * a stable array, or a final stable field via reflection or other means).
- * Since the HotSpot VM promotes a non-null component value to constant, it may
- * be that the Java memory model would appear to be broken, if such a constant
- * (the second value of the field) is used as the value of the field even after
- * the field value has changed (to a third value).
+ * After constant folding, the compiler can make use of may aspects of
+ * the object: Its dynamic type, its length (if it is an array), and
+ * the values of its fields (if they are themselves constants, either
```
liach, line 166: I wonder if this includes object identity, or the object's identity hash code - for example, published MethodType objects are all identity-based, and using their identity hash should be better than using their nominal hash code.

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
DanHeidinga: If `array` is a stable field, wouldn't `array.length` just be a (potentially compile time) constant?

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
```diff
@@ -28,56 +28,191 @@
...
+ * There are special times early in application startup when constant
+ * folding of stable variables is disabled.  Specifically, the VM does
+ * not observe the effects of any {@code @Stable} annotation until
+ * after an AOT cache assembly phase (if any) has run to completion.
+ * Therefore, during these early times, a stable variable can be
+ * changed in any way, just like a regular mutable variable (field or
+ * array component).  If a field is annotated {@code @Stable} but is
+ * treated in this way during AOT cache assembly, that fact must be
+ * clearly stated as a warning on the field declaration.  If there is
+ * no such warning, maintainers can ignore this edge case.
```
DanHeidinga, line 111: The warning on the field declaration is a comment?  As a follow on to this, should we define an additional annotation to carry that warning ie `@AssemblyTimeChangable`?  (better naming to be bike shed)

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
```diff
@@ -28,56 +28,191 @@
...
+ * evaluated data structures is to assign to stable variables much
+ * later than the enclosing data structure is created.  This means
+ * that racing reads and writes might observe nulls (or primitive
+ * zeroes) as well as non-default values.
+ * <p>
+ * Therefore, most code which reads stable variables should not assume
+ * that the value has been set, and should dynamically test for a null
+ * (or zero) value.  Code which cannot prove proper ordering of
+ * initialization may use stable variables without performing the null
+ * (or zero) test.  Code which omits the null (or zero) test should be
```
DanHeidinga, line 146:
```
 * (or zero) value.  Code which cannot prove proper ordering of
 * initialization may use stable variables without performing the null
 * (or zero) test.  Code which omits the null (or zero) test should be
```
Do we want to say "may use without null check" or is there a missing "should not" in that sentence?

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
```diff
@@ -28,56 +28,191 @@
...
+ * <em>also</em> treated as independent stable variables.  Such a
+ * stable array may contain any type, reference or primitive.
+ * <p>
+ * More generally, if a stable field is declared as an array type with
+ * <em>D</em> dimensions, then all the non-null components of the
+ * array, and of any sub-arrays up to a nesting depth less than
+ * <em>D</em>, are treated as stable variables.  Thus, a stable field
+ * declared as an array potentially defines a tree (of fixed depth
+ * <em>D</em>) containing many stable variables, with each such stable
+ * variable is independently considered for optimization.  In this
```
JornVernee, line 78:
```suggestion
 * variable independently considered for optimization.  In this
```

src/java.base/share/classes/jdk/internal/vm/annotation/Stable.java
```diff
@@ -28,56 +28,324 @@
...
+ * <em>stable variable</em>, expected to change its value just once.
+ * All Java fields are initialized by the VM with a default value
+ * (null or zero).  A properly used stable field will be set
+ * just once to a non-default value, and keep that value forever.
+ * While the field contains its default (null or zero) value,
+ * the VM treats it as an ordinary mutable variable.  When a
+ * non-default value is stored into the field, the VM is permitted
+ * to assume that no more significant changes will occur.  This in
+ * turn enables the VM to optimize uses of the stable variable, treating
+ * them as constant values.  This behavior is a useful building block
```
JornVernee, line 41: > This in turn enables the VM to optimize uses of the stable variable, treating them as constant values.

A mistake I often see people make, is thinking that `@Stable` will always allow the JIT to directly use the value of an instance field. But, this of course depends on the instance holding the field being a constant as well. I feel like this could be stated more explicitly in this text.

Maybe something along the lines of: A instance field load operation has as input an object instance, and as output the value of a particular field. There are two requirements for constant folding such a load operation: 1) the object instance is a constant. 2) the field's value will not change in the future. The `@Stable` annotation influences _only_ the second condition. It is the responsibility of the user of this annotation to make sure that a load operation on a field annotated with `@Stable` has a constant object instance as input, if constant folding of the load operation is desired as a result.

iwanowww (approved):
Looks good.

PR #28, opened 2025-01-28
8348858: [leyden] Bump the default code buffer sizes to store more generated code

shipilev:
Due to current prototype limitation, we cannot yet store the generated code that has the expanded code buffer. I tried to address that directly, but I think relocations disagree with the whole thing, so this implementation limitation stays for a bit longer. I turned the bailout due to that cause from `info` into `warning`.

On `JavacBenchApp 50`, this causes us to lose 700 (!) C2 compiled methods from the SCC! We can dodge significant part of the hit by bumping the default code buffer sizes, and thus making buffers less likely to require resizing, and thus allowing to store more code in SCC. 

Additional testing:

---------
### Progress

### Issue
 * [JDK-8348858](https://bugs.openjdk.org/browse/JDK-8348858): [leyden] Bump the default code buffer sizes to store more generated code (**Enhancement** - P4)

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - Committer) ⚠️ Review applies to [5ee9bb60](https://git.openjdk.org/leyden/pull/28/files/5ee9bb60fc0ba83d2b052b73a07b41c0b2ac8fc9)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/28/head:pull/28` \
`$ git checkout pull/28`

`$ git checkout pull/28` \
`$ git pull https://git.openjdk.org/leyden.git pull/28/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 28`

`$ git pr show -t 28`

</details>

<a href="https://git.openjdk.org/leyden/pull/28.diff">https://git.openjdk.org/leyden/pull/28.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/28#issuecomment-2619088510)
</details>

### DIFF
--- a/src/hotspot/share/opto/output.cpp
+++ b/src/hotspot/share/opto/output.cpp
@@ -27,6 +27,7 @@
 #include "code/compiledIC.hpp"
 #include "code/debugInfo.hpp"
 #include "code/debugInfoRec.hpp"
+#include "code/SCCache.hpp"
 #include "compiler/compileBroker.hpp"
 #include "compiler/compilerDirectives.hpp"
 #include "compiler/disassembler.hpp"
--- a/src/hotspot/share/opto/output.cpp
+++ b/src/hotspot/share/opto/output.cpp
@@ -1370,7 +1370,7 @@ CodeBuffer* PhaseOutput::init_buffer() {
   int exception_handler_req = HandlerImpl::size_exception_handler() + MAX_stubs_size; // add marginal slop for handler
   int deopt_handler_req     = HandlerImpl::size_deopt_handler()     + MAX_stubs_size; // add marginal slop for handler
   stub_req += MAX_stubs_size;   // ensure per-stub margin
-  code_req += MAX_inst_size;    // ensure per-instruction margin
+  code_req += max_inst_size();  // ensure per-instruction margin
 
   if (StressCodeBuffers)
     code_req = const_req = stub_req = exception_handler_req = deopt_handler_req = 0x10;  // force expansion
@@ -1567,7 +1567,7 @@ void PhaseOutput::fill_buffer(C2_MacroAssembler* masm, uint* blk_starts) {
           last_inst++;
           C->cfg()->map_node_to_block(nop, block);
           // Ensure enough space.
-          masm->code()->insts()->maybe_expand_to_ensure_remaining(MAX_inst_size);
+          masm->code()->insts()->maybe_expand_to_ensure_remaining(max_inst_size());
           if ((masm->code()->blob() == nullptr) || (!CompileBroker::should_compile_new_jobs())) {
             C->record_failure("CodeCache is full");
             return;
@@ -1696,7 +1696,7 @@ void PhaseOutput::fill_buffer(C2_MacroAssembler* masm, uint* blk_starts) {
       }
 
       // Verify that there is sufficient space remaining
-      masm->code()->insts()->maybe_expand_to_ensure_remaining(MAX_inst_size);
+      masm->code()->insts()->maybe_expand_to_ensure_remaining(max_inst_size());
       if ((masm->code()->blob() == nullptr) || (!CompileBroker::should_compile_new_jobs())) {
         C->record_failure("CodeCache is full");
         return;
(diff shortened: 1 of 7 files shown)


shipilev:
Performance tests show this realistically hits mostly G1, due to more barriers and thus more generated code, as well as broken stub size estimate.

G1:

```
# --- 32 CPUs (C2 runs mostly in background, look at "User" time)

# Before
  Time (mean ± σ):     479.8 ms ±   4.5 ms    [User: 2325.9 ms, System: 205.2 ms]
  Range (min … max):   475.2 ms … 489.4 ms    10 runs

# G1 barrier size fix
  Time (mean ± σ):     494.3 ms ±   9.0 ms    [User: 1748.9 ms, System: 208.1 ms]
  Range (min … max):   480.9 ms … 508.8 ms    10 runs

# G1 barrier size fix + adjusting sizes
  Time (mean ± σ):     472.4 ms ±   5.6 ms    [User: 876.7 ms, System: 149.5 ms]
  Range (min … max):   461.2 ms … 480.2 ms    10 runs

# --- 2 CPUs (C2 interferes with the workload more directly)

# Before
  Time (mean ± σ):     620.9 ms ±  16.9 ms    [User: 1029.8 ms, System: 120.1 ms]
  Range (min … max):   598.0 ms … 650.7 ms    10 runs

# G1 barrier size fix
  Time (mean ± σ):     591.2 ms ±  15.2 ms    [User: 960.6 ms, System: 128.9 ms]
  Range (min … max):   565.1 ms … 608.8 ms    10 runs

# G1 barrier size fix + adjusting sizes
  Time (mean ± σ):     553.1 ms ±   9.5 ms    [User: 871.9 ms, System: 127.4 ms]
  Range (min … max):   539.8 ms … 572.6 ms    10 runs
```

Parallel:


```
# --- 32 CPUs (C2 runs mostly in background, look at "User" time)

# Before
  Time (mean ± σ):     452.6 ms ±   5.6 ms    [User: 987.1 ms, System: 167.9 ms]
  Range (min … max):   444.5 ms … 459.8 ms    10 runs

# G1 barrier size fix + adjusting sizes
  Time (mean ± σ):     450.3 ms ±   3.9 ms    [User: 966.3 ms, System: 162.3 ms]
  Range (min … max):   445.5 ms … 458.9 ms    10 runs

# --- 2 CPUs (C2 interferes with the workload more directly)

# Before
  Time (mean ± σ):     540.1 ms ±  13.7 ms    [User: 835.2 ms, System: 138.9 ms]
  Range (min … max):   519.9 ms … 561.0 ms    10 runs

# G1 barrier size fix + adjusting sizes
  Time (mean ± σ):     530.7 ms ±  14.1 ms    [User: 828.5 ms, System: 134.5 ms]
  Range (min … max):   509.5 ms … 552.4 ms    10 runs
```

src/hotspot/share/opto/output.cpp
```diff
@@ -1362,8 +1362,10 @@ CodeBuffer* PhaseOutput::init_buffer() {
 
   int pad_req   = NativeCall::byte_size();
 
+  // Despite the name "stub", GC barrier stubs are emitted into
+  // the insn section, and should be counted in code_req.
   BarrierSetC2* bs = BarrierSet::barrier_set()->barrier_set_c2();
-  stub_req += bs->estimate_stub_size();
+  code_req += bs->estimate_stub_size();
```
vnkozlov, line 1368: I would be careful with this. I tried this change as part of [JDK-8331253](https://git.openjdk.org/jdk/commit/3383ad6397d5a2d8fb232ffd3e29a54e0b37b686) fix. But then I have to rollback that to avoid regression [JDK-8333226](https://github.com/openjdk/jdk/pull/19531)

It is affecting inlining decision (InlineSmallCode) and have to be handled carefully.

vnkozlov: Here is PR for original change: https://github.com/openjdk/jdk/pull/19029

vnkozlov: On other hand this line change affects only requested size and not how final instructions section size is calculated. So it could be fine.

shipilev: Dang. I have not realized this is part of inline heuristics. Let me see if I can avoid touching this part.

shipilev: Yeah, this affects inlining heuristics in a major way, like the issues you have linked described. Additionally, I realized the whole `estimate_stub_size` does not actually work well: https://github.com/openjdk/jdk/pull/23333#issuecomment-2624140301. So I am making a lateral move here, and fixing C2 stub barrier expansion code directly for Leyden.

vnkozlov (approved):
This looks reasonable.

shipilev:
I think this still does the right thing performance-wise, but I can only judge based on compilation logs. I'll see if GHA is green, and then see if we can integrate.

shipilev:
Yes, direct benchmarks show it still performs like it should. `runtime/cds` also passes well.

/integrate

PR #30, opened 2025-01-29
8348960: [leyden] compiler/c1/TestConcurrentPatching.java is stuck

shipilev:
This is seen in GHA, and reproduces well on my machine as well:

```
$ CONF=linux-x86_64-server-fastdebug make images test TEST=compiler/c1/TestConcurrentPatching.java
<stuck, timeout>
```

Test runs with `-Xcomp`. gdb "thread apply all bt" shows the compilers are idle. Supplying `-XX:-UseLockFreeCompileQueues` makes the test pass. I believe there is a bug in `UseLockFreeCompileQueues` in leyden repo. 

The comment hopefully explains what happens well. This is a corner case that seems to reproduce on the test that runs `-Xcomp` with a very few compilations.

Additional testing:
 - [ ] GHA
 - [ ] Linux x86_64 server fastdebug, `compiler/c1/TestConcurrentPatching.java`, 100x

### DIFF
--- a/src/hotspot/share/compiler/compileBroker.cpp
+++ b/src/hotspot/share/compiler/compileBroker.cpp
@@ -417,16 +417,34 @@ static bool process_pending(CompileTask* task) {
 
 void CompileQueue::transfer_pending() {
   assert(_lock->owned_by_self(), "must own lock");
-  while (!_queue.empty()) {
-    CompileTask* task = _queue.pop();
+
+  bool added_stale_tasks = false;
+  bool added_new_tasks = false;
+
+  CompileTask* task;
+  while ((task = _queue.pop()) != nullptr) {
     bool is_stale = process_pending(task);
     if (is_stale) {
       task->set_next(_first_stale);
       task->set_prev(nullptr);
       _first_stale = task;
+      added_stale_tasks = true;
     } else {
       add(task);
+      added_new_tasks = true;
     }
+  }
+
+  if (added_stale_tasks && !added_new_tasks) {
+    // We have added stale tasks. There might be waiters that want
+    // the notification these tasks have failed. Normally, this would
+    // be done by a compiler thread that would perform the purge at
+    // the end of some compilation. Butt if we never added any tasks,
+    // there is no guarantee compilers would run and do the purge.
+    // Do the purge here and now to unblock the waiters.
+    // NOTE: The call below recurses back to this method, but it should
+    // converge to having no stale tasks at all.
+    purge_stale_tasks();
   }
 }
 


shipilev:
Feel free to integrate on my behalf.

/integrate delegate

shipilev:
First GHA run is clean, use to reliably fail. Running GHA again to confirm.

shipilev:
100x runs of `compiler/c1/TestConcurrentPatching.java` are also done without timeouts. I think this confirms the fix works.

shipilev:
Ran 700x `compiler/c1/TestConcurrentPatching.java` overnight, and no timeouts. I think we are good to go.

adinn (approved):
Looks correct. However, it would be nice not to rely on recursion even though I agree it will eventually stop.

I think you can do fairly simply that by refactoring `transfer_pending`. If you relocate the while loop (pop and triage) into a separate method (`do_transfer_pending`) that you call here then you can also call that method from `purge_stale_tasks` inside a loop instead of recursing into `transfer_pending`. The new method would need to return true if a further purge is needed (i.e. `added_stale_tasks && !added_new_tasks`) so you can decide whether to call `purge_tasks` here and whether to continue looping inside `purge_tasks`.

iwanowww (approved):
Looks good. Good catch!

iwanowww:
> However, it would be nice not to rely on recursion even though I agree it will eventually stop.

IMO an optional argument in `CompileQueue::purge_stale_tasks()` to control whether `transfer_pending()` is called or not would enable an easy way to turn the recursion into a (nested) loop.

iwanowww:
Thinking more about it, does the bug arise in `CompileQueue::free_all()` due to absence of `purge_stale_task()` call? The only other place where `transfer_pending()` is used is `CompileQueue::get()`, but I don't see how it can avoid calling `purge_stale_task()`. 

Would a `purge_stale_task()` call in `CompileQueue::free_all()` fix the problem as well?

iwanowww:
Or are compiler threads simply stuck in `while (_first == nullptr) { ... }` in `CompileQueue::get()` waiting for more compilations while stale task queue remains non-empty?

Overall, I'd prefer an explicit call to `purge_stale_task()` rather than making all `transfer_pending()` calls becoming a point where MCQ lock can be released.

shipilev:
> Or are compiler threads simply stuck in `while (_first == nullptr) { ... }` in `CompileQueue::get()` waiting for more compilations while stale task queue remains non-empty?

I think this is the case: compilers are idle and never get to purge the tasks. So the "normal" lifecycle for stale task handling does not work. I understand the concern about releasing the MCQ lock in `transfer_pending`. Thinking about this, we can fix it in two ways: a) perform the purge right when the problem is constructed (this is current version); b) make sure we purge the tasks before we block the compiler thread.

For (b), a potential place that resolves some of this is before `locker.wait(5*1000);` that parks the compiler thread. We can do `purge_stale_task` before it? It will still release MCQ, but it looks fine given we are about to release it for `wait`.

Made this patch locally, and it seems to pass the test without deadlock:

```
diff --git a/src/hotspot/share/compiler/compileBroker.cpp b/src/hotspot/share/compiler/compileBroker.cpp
index 0770c1112d5..1f6ae4a8f5d 100644
--- a/src/hotspot/share/compiler/compileBroker.cpp
+++ b/src/hotspot/share/compiler/compileBroker.cpp
@@ -495,6 +495,8 @@ CompileTask* CompileQueue::get(CompilerThread* thread) {
       break;
     }
 
+    purge_stale_tasks();
+
     // If there are no compilation tasks and we can compile new jobs
     // (i.e., there is enough free space in the code cache) there is
     // no need to invoke the GC.
```

iwanowww:
I like the version with extra `purge_stale_tasks()` in `CompileQueue::get`. Releasing MCQ lock there looks benign.

shipilev:
Good. Except we also need to take care of repeated `transfer_pending` from inside of `purge_stale_tasks`, which may generate more stale tasks. And we cannot just leave pending tasks behind when we are about to block. So this IMO necessitates a bit more scaffolding. See new version. I'll run it overnight.

I also left the `is_empty` -> `pop` rewrite, because `is_empty` is racy and not supposed to be used during modifications, see the non-blocking queue docs.

shipilev:
500x runs of `TestConcurrentPatching.java` passes. `all` test suite passes without regressions (there are some Leyden-specific problems that reproduce even without this patch). I would like to integrate this version, if you agree @iwanowww?

src/hotspot/share/compiler/compileBroker.cpp
```diff
@@ -417,8 +417,9 @@ static bool process_pending(CompileTask* task) {
 
 void CompileQueue::transfer_pending() {
   assert(_lock->owned_by_self(), "must own lock");
-  while (!_queue.empty()) {
-    CompileTask* task = _queue.pop();
+
+  CompileTask* task;
```
iwanowww, line 421: I assume these changes aren't strictly needed (`_queue` is always drained under the lock).

iwanowww (approved):
Looks good.

shipilev:
Thanks!

/integrate

PR #32, opened 2025-02-05
8349457: [leyden] Workaround Leyden problems with -XX:-TieredCompilation

shipilev:
Currently, supplying `-XX:-TieredCompilation` would break with Leyden workflows, because the archive-backed code cache heap would only be created if `SegmentedCodeCache` is enabled. With default compilation policy, this happens normally when tiered compilation is enabled, but not without it.

There are other minor peculiarities that we need to handle, which are exposed with `runtime/cds` + `-XX:-TieredCompilation` runs.

This prevents us from testing C2-only Leyden configs and studying compilation dynamics without tiered policy getting in the way.

Additional testing:
 - [x] Ad-hoc Leyden experiments now work with `-XX:-TieredCompilation`: the C2-only code is being stored and loaded successfully from the CDS archive
 - [x] `runtime/cds` now passes with `-XX:-TieredCompilation`
 - [x] `runtime/cds` still passes by default (i.e. with tiered compilation)

---------
### Progress

### Issue
 * [JDK-8349457](https://bugs.openjdk.org/browse/JDK-8349457): [leyden] Workaround Leyden problems with -XX:-TieredCompilation (**Bug** - P4)

### Reviewers
 * [Ashutosh Mehra](https://openjdk.org/census#asmehra) (@ashu-mehra - Committer) ⚠️ Review applies to [1ee3b847](https://git.openjdk.org/leyden/pull/32/files/1ee3b847d231dd8ac1c9e5e6c95e2b1bac2cbc48)
 * [Igor Veresov](https://openjdk.org/census#iveresov) (@veresov - Committer) ⚠️ Review applies to [1ee3b847](https://git.openjdk.org/leyden/pull/32/files/1ee3b847d231dd8ac1c9e5e6c95e2b1bac2cbc48)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/32/head:pull/32` \
`$ git checkout pull/32`

`$ git checkout pull/32` \
`$ git pull https://git.openjdk.org/leyden.git pull/32/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 32`

`$ git pr show -t 32`

</details>

<a href="https://git.openjdk.org/leyden/pull/32.diff">https://git.openjdk.org/leyden/pull/32.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/32#issuecomment-2636586188)
</details>


src/hotspot/share/compiler/precompiler.cpp
```diff
@@ -232,16 +234,24 @@ void Precompiler::compile_cached_code(ArchiveBuilder* builder, TRAPS) {
...
-        comp_level = CompLevel_limited_profile;
-      }
-      PrecompileIterator pi(comp_level, false /*for_preload*/, (CompLevel)level, THREAD);
+    {
+      PrecompileIterator pi(CompLevel_full_optimization, false /*for_preload*/, CompLevel_full_optimization, THREAD);
       TrainingData::iterate(pi);
       pi.precompile(builder, THREAD);
     }
 
+    if (TieredCompilation) {
```
adinn, line 243: How does this work as far as consistency between training/assembly run and production run is concerned? Do we need to enforce the same setting across these runs or can we e.g. train and the (assuming we fork) assemble with -TieredCompilation but production run with +TieredCompilation -- or vice versa?

shipilev: I tested different combinations of `{+|-}TieredCompilation` with training and production runs, and it works well. I believe the normal mechanisms make it right. No C1 code? Nothing loaded. C1 code in archive, but C1 is disabled? We are not loading it then. 

In fact, there seem to be a slight performance edge in training with C2 only, and then running with C1+C2: no penalty on loading lots of C1 compiled code, and quicker transition from intepreter to "real" C1 in only a few places where it is needed! Compiler dynamics is weird.

ashu-mehra (approved):
lgtm

vnkozlov (commented):
@veresov should look on this.

veresov:
> @veresov should look on this.

@iwanowww wrote that particular part. I looks good to me though.

shipilev:
Fun perf results:

```
$ build/linux-x86_64-server-release/images/jdk/bin/java -Xms64m -Xmx1g -XX:CacheDataStore=JavacBenchApp.cds -XX:+UseParallelGC -cp JavacBenchApp.jar JavacBenchApp 50

# --- Training with C1 and C2 (+Tiered)

# Production with C1 and C2 (+Tiered)
  Time (mean ± σ):     434.0 ms ±   8.5 ms    [User: 1506.7 ms, System: 218.4 ms]
  Range (min … max):   423.4 ms … 449.2 ms    10 runs

# Production with C2 only (-Tiered)
  Time (mean ± σ):     561.2 ms ±  12.8 ms    [User: 3578.0 ms, System: 263.2 ms]
  Range (min … max):   542.3 ms … 589.2 ms    10 runs

# --- Training with C2 only (-Tiered)

# Production with C1 and C2 (+Tiered)
  Time (mean ± σ):     408.0 ms ±   5.4 ms    [User: 1183.2 ms, System: 165.2 ms]
  Range (min … max):   401.3 ms … 421.0 ms    10 runs

# Production with C2 only (-Tiered)
  Time (mean ± σ):     482.0 ms ±  11.6 ms    [User: 1959.1 ms, System: 182.5 ms]
  Range (min … max):   467.2 ms … 508.2 ms    10 runs

```

shipilev:
Feel free to delegate without me.

/integrate delegate

src/hotspot/share/compiler/precompiler.cpp
iwanowww: As I mentioned during the meeting, it looks weird to me to see `if (TieredCompilation)` checks in Precompiler code which is executed only during assembly phase. As a data point, regular CDS archive is used to be generated in interpreter-only mode (to make the process reproducible). 

Instead of supporting arbitrary JVM configurations during assembly phase, I'd prefer to see offending JVM parameters being overridden/dropped.

shipilev: Right, I thought it matters for Precompiler not to attempt pre-compiling C1 classes with `-XX:-TieredCompilation`, but it does not. I see `runtime/cds` passing with/without TieredCompilation without these hunks, and CDS archive sizes are similar, and performance is still the same. So I have reverted Precompiler hunks as unnecessary.

Take a look again?

shipilev:

src/hotspot/share/compiler/precompiler.cpp
ashu-mehra: > CDS archive sizes are similar,

I think this would impact the code archive, not the CDS archive. 
If it does reduce the on-disk size of code archive, then why not do this? It's pretty straightforward; not a complicated change IMO.

shipilev: Do you see the same kind of improvement with new version of the patch and `-XX:-TieredCompilation`?

ashu-mehra: > Do you see the same kind of improvement with new version of the patch and -XX:-TieredCompilation?

aah! I get it now. For some reason I kept thinking we are not able to run with `-XX:-TieredCompilation` in the training/assembly run and this particular hunk fixes it. But that's not the case. My bad!

shipilev: No problem! The interactions between different phases and compiler tiers are often confusing :)

iwanowww (approved):
Looks good.

shipilev:
Thanks!

/integrate

PR #33, opened 2025-02-06
8349539: [leyden] Allow dumping nmethods that reference archived hidden klasses

shipilev:
Studying the bailout reasons for JavacBenchApp, I notices the majority of nmethod store bailouts are due to referencing the hidden klasses. Those checks seem to be from the initial commit. I wonder if we can actually skip hidden klass checks, now that we archive lots of those since invokedynamic AOT was implemented.

---------
### Progress

### Issue
 * [JDK-8349539](https://bugs.openjdk.org/browse/JDK-8349539): [leyden] Allow dumping nmethods that reference archived hidden klasses (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/33/head:pull/33` \
`$ git checkout pull/33`

`$ git checkout pull/33` \
`$ git pull https://git.openjdk.org/leyden.git pull/33/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 33`

`$ git pr show -t 33`

</details>

<a href="https://git.openjdk.org/leyden/pull/33.diff">https://git.openjdk.org/leyden/pull/33.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/33#issuecomment-2639624243)
</details>


shipilev:
Look at "user" time, this is what we save without recompiling the methods that get bailed from the archive.

```
$ rm -f *.cds*; hyperfine -w 3 -r 10 "build/linux-x86_64-server-release/images/jdk/bin/java -XX:CacheDataStore=JavacBenchApp.cds -cp JavacBenchApp.jar JavacBenchApp 500";

# Before
  Time (mean ± σ):      1.556 s ±  0.013 s    [User: 4.575 s, System: 0.501 s]
  Range (min … max):    1.538 s …  1.580 s    10 runs

# After
  Time (mean ± σ):      1.551 s ±  0.016 s    [User: 3.646 s, System: 0.458 s]
  Range (min … max):    1.536 s …  1.586 s    10 runs
```

iwanowww (commented):
I agree that the restriction for hidden classes can be lifted.

But there are non-default modes where it still matters (`UseMetadataPointers == true`). 
I suggest to keep `klass->is_hidden()` check, but move it down to guard symbolic resolution case.

vnkozlov:
>>But there are non-default modes where it still matters (UseMetadataPointers == true).
>FWIW, I believe UseMetadataPointers is true by default. Turning it off starts to fail runtime/cds.

I started implementation of Code caching using only symbolic resolution in relocations. When it is integrated with other changes I left such capability guarded by this flag.
We can retire this flag when the only way to use code caching is CDS archive which will contain the code instead of separate file. I am currently looking on this.

shipilev:

shipilev:
Are we good with this, @iwanowww? Is this the correct place to move the check?

iwanowww (approved):
Looks good.

shipilev:
Thanks!

/integrate

PR #34, opened 2025-02-10
8349713: [leyden] Memory map the cached code file

shipilev:
It is visible in profiles for lots of applications that reading the SC cache file at startup costs significantly. On JavacBenchApp example, loading ~25M code requires about 30ms. This is ~1 GB/sec, so it is I/O limited.

We should really mmap the SC cache file to alleviate these costs. Let the actual SC readers (separate threads) to eat the cost of reading from the backing file. 

I was not entirely sure COW for file mappings works correctly on Windows, so I excepted that one.

Additional testing:

---------
### Progress

### Issue
 * [JDK-8349713](https://bugs.openjdk.org/browse/JDK-8349713): [leyden] Memory map the cached code file (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/34/head:pull/34` \
`$ git checkout pull/34`

`$ git checkout pull/34` \
`$ git pull https://git.openjdk.org/leyden.git pull/34/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 34`

`$ git pr show -t 34`

</details>

<a href="https://git.openjdk.org/leyden/pull/34.diff">https://git.openjdk.org/leyden/pull/34.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/34#issuecomment-2647824324)
</details>

### DIFF
--- a/src/hotspot/share/code/SCCache.cpp
+++ b/src/hotspot/share/code/SCCache.cpp
@@ -549,19 +549,28 @@ SCCache::SCCache(const char* cache_path, int fd, uint load_size) {
 
   _use_meta_ptrs = UseSharedSpaces ? UseMetadataPointers : false;
 
-  // Read header at the begining of cache
-  uint header_size = sizeof(SCCHeader);
   if (_for_read) {
     // Read cache
-    _C_load_buffer = NEW_C_HEAP_ARRAY(char, load_size + DATA_ALIGNMENT, mtCode);
-    _load_buffer = align_up(_C_load_buffer, DATA_ALIGNMENT);
-    uint n = (uint)::read(fd, _load_buffer, load_size);
-    if (n != load_size) {
-      log_warning(scc, init)("Failed to read %d bytes at address " INTPTR_FORMAT " from Startup Code Cache file '%s'", load_size, p2i(_load_buffer), _cache_path);
+    if (MmapCachedCode) {
+      // Implementation note: the mapping is not read-only, because entries are updated in-place.
+      // However, this mapping is MAP_PRIVATE, which means the changes to mapping do not reflect
+      // in the backing file. TODO: Avoid in-place modifications, so we do not have to rely on this?
+      _load_buffer = os::map_memory(fd, _cache_path, 0, nullptr, load_size,
+                                   false,false, mtCode);
+    } else {
+      _C_load_buffer = NEW_C_HEAP_ARRAY(char, load_size + DATA_ALIGNMENT, mtCode);
+      _load_buffer = align_up(_C_load_buffer, DATA_ALIGNMENT);
+      uint n = (uint)::read(fd, _load_buffer, load_size);
+      if (n != load_size) {
+        _load_buffer = nullptr;
+      }
+    }
+    if (_load_buffer == nullptr) {
+      log_warning(scc, init)("Failed to read/mmap %d bytes at address " INTPTR_FORMAT " from Startup Code Cache file '%s'", load_size, p2i(_load_buffer), _cache_path);
       set_failed();
       return;
     }
-    log_info(scc, init)("Read %d bytes at address " INTPTR_FORMAT " from Startup Code Cache '%s'", load_size, p2i(_load_buffer), _cache_path);
+    log_info(scc, init)("Read/mmaped %d bytes at address " INTPTR_FORMAT " from Startup Code Cache '%s'", load_size, p2i(_load_buffer), _cache_path);
 
     _load_header = (SCCHeader*)addr(0);
     const char* scc_jvm_version = addr(_load_header->jvm_version_offset());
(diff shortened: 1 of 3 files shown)


shipilev:
It demonstrably improves performance on Linux:

```
# Without mmap (legacy code)
Benchmark 1: build/linux-x86_64-server-release/images/jdk/bin/java -Xms64m -Xmx1g -XX:CacheDataStore=JavacBenchApp.cds -XX:+UseParallelGC -cp JavacBenchApp.jar -XX:-MmapCachedCode JavacBenchApp 50 1
  Time (mean ± σ):     408.0 ms ±   2.5 ms    [User: 1231.7 ms, System: 196.2 ms]
  Range (min … max):   404.6 ms … 412.8 ms    10 runs

# With mmap
Benchmark 1: build/linux-x86_64-server-release/images/jdk/bin/java -Xms64m -Xmx1g -XX:CacheDataStore=JavacBenchApp.cds -XX:+UseParallelGC -cp JavacBenchApp.jar JavacBenchApp 50 1
  Time (mean ± σ):     382.1 ms ±   2.6 ms    [User: 1229.9 ms, System: 181.6 ms]
  Range (min … max):   378.9 ms … 388.0 ms    10 runs
```

franz1981:
@shipilev 

Does the numbers still holds with
```
sync; echo 3 > /proc/sys/vm/drop_caches
```
before the benchmark (single shot) run?
With big servers, unless there's some expected sharing (i.e. multiple processes using the same archive) to boost it, is that direct I/O could still benefit loading the archive by saving the extra copy (in the OS page cache) required to use the data read.
The other concern re mmap is due to munmap cost which, on kernel side, relies (IIRC) to some v page single (!) lock to guard it - which usually slowdown processes termination

shipilev:
> Does the numbers still holds with sync; echo 3 > /proc/sys/vm/drop_caches

Yes, they do, and there is a good reason why: without caches, the hit on the critical startup path is even worse, even with a modern SSD. And AFAIU, file-backed mmap does play well with I/O caches too. Observe:

```
# Drop caches, read
  Time (mean ± σ):     521.0 ms ±   6.1 ms    [User: 1297.3 ms, System: 259.0 ms]
  Range (min … max):   514.6 ms … 532.2 ms    10 runs

# Drop caches, mmap
  Time (mean ± σ):     479.9 ms ±   2.6 ms    [User: 1148.7 ms, System: 223.4 ms] ; <--- ~40ms faster
  Range (min … max):   476.4 ms … 484.3 ms    10 runs

# Cached, read
  Time (mean ± σ):     413.0 ms ±   3.5 ms    [User: 1267.0 ms, System: 207.9 ms]
  Range (min … max):   408.6 ms … 417.7 ms    10 runs

# Cached, mmap
  Time (mean ± σ):     386.0 ms ±   4.7 ms    [User: 1258.5 ms, System: 183.3 ms]  ; <--- ~30ms faster
  Range (min … max):   378.7 ms … 393.7 ms    10 runs
```

shipilev:

shipilev:
> Good. Until, as we discussed, cached code (and its data) will be part of CDS archive file.

Yes, that would make this whole thing no-op. CDS mmap-s the archive already.

> How CDS handles `mmap` on windows?

Windows can still memory-map the file through its Windows APIs. But I _vaguely_ recollect some corner cases that @iklam fights every so often. Something with remapping or so? See `MetaspaceShared::use_windows_memory_mapping`, for example. I don't think it is wise to spend time dealing with those corner cases in current Leyden prototype, let it be part of whole CDS archive first.

iklam:
> > Good. Until, as we discussed, cached code (and its data) will be part of CDS archive file.
> 
> Yes, that would make this whole thing no-op. CDS mmap-s the archive already.
> 
> > How CDS handles `mmap` on windows?
> 
> Windows can still memory-map the file through its Windows APIs. But I _vaguely_ recollect some corner cases that @iklam fights every so often in CDS. Something with remapping or so? See `MetaspaceShared::use_windows_memory_mapping`, for example. I don't think it is wise to spend time dealing with those corner cases in current Leyden prototype, let it be part of whole CDS archive first.

With the Windows APIs used by HotSpot today (VirtualAlloc and MapViewOfFile), we can't map a file into a reserved region. If we just want to mmap the SCCache into non-reserved, random location that's picked by the OS, we can already do that today with HotSpot.

In some cases, CDS wants to mmap into reserved regions. On Windows, we end up not mapping with MapViewOfFile, but simply reading the entire CDS file into reserved memory.

I think with this new API [MapViewOfFile3](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile3), we can map into a reserved region. This would be useful if, for example, we want the SCCache to be immediately next to CDS, so that the AOT code can use relative addressing for metadata pointers (InstanceKlass*, etc).

shipilev:
> With the Windows APIs used by HotSpot today (VirtualAlloc and MapViewOfFile), we can't map a file into a reserved region. If we just want to mmap the SCCache into non-reserved, random location that's picked by the OS, we can already do that today with HotSpot.
> 
> In some cases, CDS wants to mmap into reserved regions. On Windows, we end up not mapping with MapViewOfFile, but simply reading the entire CDS file into reserved memory.

OK, great. AFAICS, this technically allows us to do `MmapCachedCode = true` on Windows too. Non-Windows currently rely on `MAP_PRIVATE` to get us COW. I believe `FILE_MAP_COPY` in `os::pd_map_memory` gives us the same on Windows.

Still, I think it is a bit saner to keep doing mmap-ing only on Linux, if only to test that non-mmap path works and not force us to debug Windows mmap issues in Leyden prototype. But I don't feel strongly about this. Opinions welcome!

shipilev:
Any other opinions about this? I would like to integrate this to reap some startup benefits :)

vnkozlov:
> I think this looks fine. The change is small, so we can update or remove it when migrating the SCCache into CDS.
> 
> @vnkozlov what do you think?

Yes, that is the plan.

shipilev:
I think this PR would be superseded by @ashu-mehra's #39. We can still do it ahead of #39, removing the flag and accepting all platforms.

ashu-mehra:
@shipilev I am fine with merging this as is.

PR #39, opened 2025-02-11
Store AOT code in cached code region of AOT cache

ashu-mehra:
This PR stores AOT code in the cached code region of AOT Cache.

Some things worth pointing out:
1. Before this patch: the AOT code is first generated in a temporary buffer. After all the code is generated, this temporary buffer is copied into second buffer (allocated in native heap) for multiple reasons. This second buffer is then stored in the file as is. With this patch the second buffer is now allocated in the cached code region which is part of the AOT cache. So now the cached code region contains the permanent oops table (which already exists before this patch) and the AOT code+data.
2. I have removed `_jvm_version_offset` from the SCCHeader. AOT Cache already stores VM identifier which should cover it.
~3. I have to remove the code that maps cached code region to CodeCache because the contents in cached code region are not CodeBlobs. The cached code region can be mapped to any random address.~
4. I have replaced references to "Startup Code Cache" with "AOT Code Cache" or "AOT Code".

---------
### Progress

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - Committer) ⚠️ Review applies to [a63db458](https://git.openjdk.org/leyden/pull/39/files/a63db4580a5a9ea5ecf8d1de3030784e70472f91)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/39/head:pull/39` \
`$ git checkout pull/39`

`$ git checkout pull/39` \
`$ git pull https://git.openjdk.org/leyden.git pull/39/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 39`

`$ git pr show -t 39`

</details>

<a href="https://git.openjdk.org/leyden/pull/39.diff">https://git.openjdk.org/leyden/pull/39.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/39#issuecomment-2652193725)
</details>

### DIFF
--- a/src/hotspot/share/cds/archiveBuilder.cpp
+++ b/src/hotspot/share/cds/archiveBuilder.cpp
@@ -312,7 +312,7 @@ void ArchiveBuilder::sort_klasses() {
 }
 
 address ArchiveBuilder::reserve_buffer() {
-  size_t buffer_size = LP64_ONLY(CompressedClassSpaceSize) NOT_LP64(256 * M);
+  size_t buffer_size = LP64_ONLY(CompressedClassSpaceSize) NOT_LP64(256 * M) + ReservedCodeCacheSize;
   ReservedSpace rs = MemoryReserver::reserve(buffer_size,
                                              MetaspaceShared::core_region_alignment(),
                                              os::vm_page_size());
--- a/src/hotspot/share/cds/archiveBuilder.cpp
+++ b/src/hotspot/share/cds/archiveBuilder.cpp
@@ -312,7 +312,8 @@ void ArchiveBuilder::sort_klasses() {
 }
 
 address ArchiveBuilder::reserve_buffer() {
-  size_t buffer_size = LP64_ONLY(CompressedClassSpaceSize) NOT_LP64(256 * M) + ReservedCodeCacheSize;
+  // CachedCodeMaxSize accounts for cached code region.
+  size_t buffer_size = LP64_ONLY(CompressedClassSpaceSize) NOT_LP64(256 * M) + CachedCodeMaxSize;
   ReservedSpace rs = MemoryReserver::reserve(buffer_size,
                                              MetaspaceShared::core_region_alignment(),
                                              os::vm_page_size());
--- a/src/hotspot/share/cds/archiveBuilder.cpp
+++ b/src/hotspot/share/cds/archiveBuilder.cpp
@@ -43,6 +43,7 @@
 #include "classfile/symbolTable.hpp"
 #include "classfile/systemDictionaryShared.hpp"
 #include "classfile/vmClasses.hpp"
+#include "code/SCCache.hpp"
 #include "interpreter/abstractInterpreter.hpp"
 #include "jvm.h"
 #include "logging/log.hpp"
@@ -312,8 +313,8 @@ void ArchiveBuilder::sort_klasses() {
 }
 
 address ArchiveBuilder::reserve_buffer() {
-  // CachedCodeMaxSize accounts for cached code region.
-  size_t buffer_size = LP64_ONLY(CompressedClassSpaceSize) NOT_LP64(256 * M) + CachedCodeMaxSize;
+  // SCCache::max_aot_code_size() accounts for cached code region.
+  size_t buffer_size = LP64_ONLY(CompressedClassSpaceSize) NOT_LP64(256 * M) + SCCache::max_aot_code_size();
   ReservedSpace rs = MemoryReserver::reserve(buffer_size,
                                              MetaspaceShared::core_region_alignment(),
                                              os::vm_page_size());
--- a/src/hotspot/share/cds/cdsAccess.hpp
+++ b/src/hotspot/share/cds/cdsAccess.hpp
@@ -61,6 +61,7 @@ public:
   static size_t get_cached_code_size() NOT_CDS_JAVA_HEAP_RETURN_(0);
   static void set_cached_code_size(size_t sz) NOT_CDS_JAVA_HEAP_RETURN;
 
+  static bool is_cached_code_region_empty() NOT_CDS_JAVA_HEAP_RETURN_(true);
   static bool map_cached_code(ReservedSpace rs) NOT_CDS_JAVA_HEAP_RETURN_(false);
 
   template <typename T>
--- a/src/hotspot/share/cds/cdsAccess.hpp
+++ b/src/hotspot/share/cds/cdsAccess.hpp
@@ -56,7 +56,7 @@ public:
 
   static void test_heap_access_api() NOT_CDS_JAVA_HEAP_RETURN;
 
-  static void* allocate_from_code_cache(size_t size) NOT_CDS_RETURN_(nullptr);
+  static void* allocate_from_code_cache(size_t size) NOT_CDS_JAVA_HEAP_RETURN_(nullptr);
 
   static size_t get_cached_code_size() NOT_CDS_JAVA_HEAP_RETURN_(0);
   static void set_cached_code_size(size_t sz) NOT_CDS_JAVA_HEAP_RETURN;
--- a/src/hotspot/share/cds/cdsAccess.hpp
+++ b/src/hotspot/share/cds/cdsAccess.hpp
@@ -56,13 +56,13 @@ public:
 
   static void test_heap_access_api() NOT_CDS_JAVA_HEAP_RETURN;
 
-  static void* allocate_from_code_cache(size_t size) NOT_CDS_JAVA_HEAP_RETURN_(nullptr);
+  static void* allocate_from_code_cache(size_t size) NOT_CDS_RETURN_(nullptr);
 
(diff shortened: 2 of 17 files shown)


ashu-mehra:
> I have to remove the code that maps cached code region to CodeCache because the contents in cached code region are not CodeBlobs. The cached code region can be mapped to any random address.

hmm, now I am not sure if this is really required. We can map the cached code region to CodeCache and still use it in SCCache.

vnkozlov:
> > I have to remove the code that maps cached code region to CodeCache because the contents in cached code region are not CodeBlobs. The cached code region can be mapped to any random address.
> 
> hmm, now I am not sure if this is really required. We can map the cached code region to CodeCache and still use it in SCCache.

Can we leave this part of code but not use it - map to random address for these changes?

vnkozlov:
It was prepared for "preload" AOT nmethods so we can use them immediately (after your work on caching whole nmethod). Normal AOT nmethods (tier1-4) may live in separate space until they are requested - we can't publish (patch) them until corresponding classes are initialized.

ashu-mehra:
> Can we leave this part of code but not use it - map to random address for these changes?

Yes, we can do that. It would also mean the ReservedSpace created by CodeCache would just be a lot bigger now by the size of the cached code region. But again it would only affect the virtual memory, as we are not mapping anything to that space.

vnkozlov:
I am straggling with Minimal and Zero build too for may changes :(

ashu-mehra:
Minimal is tricky. Now SCCache depends on CDS archive, so ideally its functionality should be guarded by INCLUDE_CDS and other CDS related macros.

shipilev:
> Minimal is tricky. Now SCCache depends on CDS archive, so ideally its functionality should be guarded by INCLUDE_CDS and other CDS related macros.

Minimal is easy enough to fix, we just need to clearly introduce dependency on CDS, see #40.

src/hotspot/share/cds/archiveBuilder.cpp
```diff
@@ -312,7 +312,7 @@ void ArchiveBuilder::sort_klasses() {
 }
 
 address ArchiveBuilder::reserve_buffer() {
-  size_t buffer_size = LP64_ONLY(CompressedClassSpaceSize) NOT_LP64(256 * M);
+  size_t buffer_size = LP64_ONLY(CompressedClassSpaceSize) NOT_LP64(256 * M) + ReservedCodeCacheSize;
```
vnkozlov, line 315: Please add comment about `ReservedCodeCacheSize`. Why not `CachedCodeMaxSize`?

src/hotspot/share/code/codeCache.cpp
```diff
@@ -356,6 +356,7 @@ void CodeCache::initialize_heaps() {
   }
 }
 
+
```
vnkozlov, line 359: You can undo that to avoid changing file.

test/hotspot/jtreg/premain/lib/DemoSupport.gmk
```diff
@@ -160,7 +160,7 @@ ${DEMO_CDS}: ${DEMO_JAR}
 	    -Xlog:scc,cds=debug,cds+class=debug,cds+heap=warning,cds+resolve=debug:file=${DEMO_CDS}.log::filesize=0 \
 	     ${DEMO_CMDLINE}
 	ls -l ${DEMO_CDS}
-	ls -l ${DEMO_CDS}.code
+	#ls -l ${DEMO_CDS}.code
```
vnkozlov, line 163: I would just remove this line.

src/hotspot/share/cds/archiveBuilder.cpp
ashu-mehra: > Why not CachedCodeMaxSize?

Yeah, I think that makes sense. Since the aot code buffer is based on CachedCodeMaxSize, we can use this for sizing cached code region as well.

test/hotspot/jtreg/premain/lib/DemoSupport.gmk
ashu-mehra: Done

src/hotspot/share/code/codeCache.cpp
ashu-mehra: Done

src/hotspot/share/code/SCCache.cpp
```diff
@@ -551,39 +462,31 @@ SCCache::SCCache(const char* cache_path, int fd, uint load_size) {
 
   if (_for_read) {
     // Read cache
-    if (MmapCachedCode) {
```
shipilev, line 554: I think you can drop `MmapCachedCode` flag as well.

ashu-mehra: right

src/hotspot/share/code/SCCache.cpp
```diff
@@ -1147,17 +1035,18 @@ bool SCCache::finish_write() {
     uint preload_entries_size = code_count * sizeof(uint);
     // _write_position should include code and strings
     uint code_alignment = code_count * DATA_ALIGNMENT; // We align_up code size when storing it.
-    uint total_size = _write_position + _load_size + header_size + vm_version_size +
+    uint total_size = _write_position + _load_size + header_size +
                      code_alignment + search_size + preload_entries_size + entries_size;
 
+    assert(total_size < ReservedCodeCacheSize, "Cached code region size (" UINT32_FORMAT " bytes) in AOT Cache is less than the required size (" UINT32_FORMAT " bytes).",
+           total_size, (uint)ReservedCodeCacheSize);
+
```
vnkozlov, line 1043: Should you compare to `CachedCodeMaxSize`? Also missing `Code` in "AOT Cache".

src/hotspot/share/code/SCCache.cpp
```diff
@@ -163,7 +159,7 @@ void SCCache::init2() {
     address byte_map_base = ci_card_table_address_as<address>();
     if (is_on_for_write() && !external_word_Relocation::can_be_relocated(byte_map_base)) {
       // Bail out since we can't encode card table base address with relocation
-      log_warning(scc, init)("Can't create Startup Code Cache because card table base address is not relocatable: " INTPTR_FORMAT, p2i(byte_map_base));
+      log_warning(scc, init)("Can't create Code Cache because card table base address is not relocatable: " INTPTR_FORMAT, p2i(byte_map_base));
```
vnkozlov, line 162: Missing "AOT" in "Code Cache"

vnkozlov (commented):
Did you measure any performance with this change?

vnkozlov:
Now we have 32-bits ARM build issue.

ashu-mehra:
> Now we have 32-bits ARM build issue.

yeah, I think I know what the issue is here. Testing the fix.

vnkozlov (approved):
Looks good now.

src/hotspot/share/cds/cdsAccess.hpp
```diff
@@ -56,11 +56,12 @@ class CDSAccess : AllStatic {
 
   static void test_heap_access_api() NOT_CDS_JAVA_HEAP_RETURN;
 
-  static void* allocate_from_code_cache(size_t size) NOT_CDS_RETURN_(nullptr);
+  static void* allocate_from_code_cache(size_t size) NOT_CDS_JAVA_HEAP_RETURN_(nullptr);
```
iklam, line 59: I think the following functions should be marked as `NOT_CDS_RETURN`. In the cpp file, they should be moved outside of the `INCLUDE_CDS_JAVA_HEAP` block:

```
void* CDSAccess::allocate_from_code_cache(size_t size)
size_t CDSAccess::get_cached_code_size()
void CDSAccess::set_cached_code_size(size_t sz)
bool CDSAccess::is_cached_code_region_empty()
bool CDSAccess::map_cached_code(ReservedSpace rs)
```

src/hotspot/share/code/SCCache.cpp
```diff
@@ -380,59 +381,29 @@ bool SCCache::allow_const_field(ciConstant& value) {
...
+
+  void set_aot_code_data(uint size, char* aot_data) {
+    _aot_code_size = size;
+    CDSAccess::set_pointer(&_aot_code_data, aot_data);
+  }
+
+  static CachedCodeDirectory* create();
 };
 
 // Skeleton code for including cached code in CDS:
```
iklam, line 409: The comment "Skeleton code for including cached code in CDS" should be changed to something like "Storing AOT code in CDS"

iklam (commented):
Looks good overall. Some minor comments.

Also, I think you should remove `aotFile` from test/lib/jdk/test/lib/cds/CDSAppTester.java

src/hotspot/share/cds/cdsAccess.hpp
ashu-mehra: yup, I agree

iklam (approved):
LGTM

ashu-mehra:
Thanks @shipilev @iklam @vnkozlov for the help and reviews.

PR #38, opened 2025-02-11
8349814: [leyden] Reduce uncommon traps in preload code

shipilev:
Preload code is the C2 code we load very early, before it gets substituted by more advanced C2 code. We are seeing lots of uncommon traps from the preloaded code, though, which invalidates it too early. We can do precompile code without uncommon traps to alleviate this. 

Additional testing:
 - [x] Ad-hoc perf tests (see comment below)
 - [x] Linux x86_64 server fastdebug, `runtime/cds`

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Issue
 * [JDK-8349814](https://bugs.openjdk.org/browse/JDK-8349814): [leyden] Reduce uncommon traps in preload code (**Enhancement** - P4)

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - Committer) ⚠️ Review applies to [4051ce11](https://git.openjdk.org/leyden/pull/38/files/4051ce1190dc3d6d4876276b67201515810ce174)
 * [Vladimir Ivanov](https://openjdk.org/census#vlivanov) (@iwanowww - Committer)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/38/head:pull/38` \
`$ git checkout pull/38`

Update a local copy of the PR: \
`$ git checkout pull/38` \
`$ git pull https://git.openjdk.org/leyden.git pull/38/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 38`

View PR using the GUI difftool: \
`$ git pr show -t 38`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/leyden/pull/38.diff">https://git.openjdk.org/leyden/pull/38.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/38#issuecomment-2650448615)
</details>

PR #44, opened 2025-02-21
8350488: [leyden] Experimental AOT-only mode

shipilev:
There are interesting use cases where we want the AOT-only mode. We can emulate this in current Leyden prototype by relying on preload code, and stopping any profiling, which would naturally lead to no JIT compilations. This would also make interpreter code a bit faster in case we need to fall back there. This mode looks also helpful for studying the compiler dynamics.

Additional testing:
 - [x] Eyeballing compilation logs with `-XX:+PreloadOnly`
 - [x] Linux x86_64 server fastdebug, `runtime/cds` with `-XX:+PreloadOnly`

---------
### Progress

### Issue
 * [JDK-8350488](https://bugs.openjdk.org/browse/JDK-8350488): [leyden] Experimental AOT-only mode (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/44/head:pull/44` \
`$ git checkout pull/44`

`$ git checkout pull/44` \
`$ git pull https://git.openjdk.org/leyden.git pull/44/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 44`

`$ git pr show -t 44`

</details>

<a href="https://git.openjdk.org/leyden/pull/44.diff">https://git.openjdk.org/leyden/pull/44.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/44#issuecomment-2674201600)
</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp
@@ -1186,7 +1186,7 @@ void TemplateInterpreterGenerator::bang_stack_shadow_pages(bool native_call) {
 // native method than the typical interpreter frame setup.
 address TemplateInterpreterGenerator::generate_native_entry(bool synchronized, bool runtime_upcalls) {
   // determine code generation flags
-  bool inc_counter  = UseCompiler || CountCompiledCalls;
+  bool inc_counter = (UseCompiler || CountCompiledCalls) && !PreloadOnly;
 
   // r1: Method*
   // rscratch1: sender sp
@@ -1638,7 +1638,7 @@ address TemplateInterpreterGenerator::generate_native_entry(bool synchronized, b
 //
 address TemplateInterpreterGenerator::generate_normal_entry(bool synchronized, bool runtime_upcalls) {
   // determine code generation flags
-  bool inc_counter  = UseCompiler || CountCompiledCalls;
+  bool inc_counter = (UseCompiler || CountCompiledCalls) && !PreloadOnly;
 
   // rscratch1: sender sp
   address entry_point = __ pc();
--- a/src/hotspot/cpu/arm/templateInterpreterGenerator_arm.cpp
+++ b/src/hotspot/cpu/arm/templateInterpreterGenerator_arm.cpp
@@ -809,7 +809,7 @@ address TemplateInterpreterGenerator::generate_Float_floatToFloat16_entry() { re
 
 address TemplateInterpreterGenerator::generate_native_entry(bool synchronized, bool runtime_upcalls) {
   // determine code generation flags
-  bool inc_counter  = UseCompiler || CountCompiledCalls;
+  bool inc_counter = (UseCompiler || CountCompiledCalls) && !PreloadOnly;
 
   // Incoming registers:
   //
@@ -1143,7 +1143,7 @@ address TemplateInterpreterGenerator::generate_native_entry(bool synchronized, b
 //
 address TemplateInterpreterGenerator::generate_normal_entry(bool synchronized, bool runtime_upcalls) {
   // determine code generation flags
-  bool inc_counter  = UseCompiler || CountCompiledCalls;
+  bool inc_counter = (UseCompiler || CountCompiledCalls) && !PreloadOnly;
 
   // Rmethod: Method*
   // Rthread: thread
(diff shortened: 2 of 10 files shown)


shipilev:
Sample performance results, after C2-only pre-training `JavacBenchApp` with 200x iterations:

```
# --- 32 cores

# Default
  Time (mean ± σ):     397.2 ms ±   8.3 ms    [User: 1015.7 ms, System: 160.4 ms]
  Range (min … max):   388.7 ms … 434.4 ms    30 runs
 
# -XX:+PreloadOnly
  Time (mean ± σ):     414.6 ms ±   8.0 ms    [User: 509.2 ms, System: 78.6 ms]
  Range (min … max):   400.8 ms … 447.4 ms    30 runs
 
# --- 2 cores

# Default
  Time (mean ± σ):     528.3 ms ±  8.6 ms    [User: 866.2 ms, System: 111.4 ms]
  Range (min … max):   472.5 ms … 577.8 ms    30 runs
 
# -XX:+PreloadOnly
  Time (mean ± σ):     405.0 ms ±   5.1 ms    [User: 495.7 ms, System: 71.6 ms]
  Range (min … max):   397.4 ms … 416.5 ms    30 runs
```

In both cases, "user" time goes down because we have no JIT compilations. In 32 core case, we can see that peak performance suffers a bit, since preload generated code is not 100% efficient. The combination of these two factors is a net benefit in 2 core case: not doing JIT compilations more than pays for preload code inefficiency. This tradeoff of course depends on how well-trained the scenario is.

shipilev:
Sample performance results, after pre-training `JavacBenchApp` with 200x iterations:

```
# === 32 cores

# Default
  Time (mean ± σ):     403.4 ms ±   6.5 ms    [User: 1082.1 ms, System: 167.6 ms]
  Range (min … max):   387.6 ms … 417.7 ms    30 runs

 
# -XX:+PreloadOnly
  Time (mean ± σ):     439.1 ms ±   5.3 ms    [User: 544.2 ms, System: 84.2 ms]
  Range (min … max):   430.3 ms … 456.2 ms    30 runs

 
# === 2 cores

# Default
  Time (mean ± σ):     531.8 ms ±  30.1 ms    [User: 870.9 ms, System: 114.3 ms]
  Range (min … max):   479.6 ms … 606.8 ms    30 runs
 
# -XX:+PreloadOnly
  Time (mean ± σ):     425.8 ms ±   6.4 ms    [User: 530.2 ms, System: 76.4 ms]
  Range (min … max):   418.9 ms … 451.4 ms    30 runs
```

In both cases, "user" time goes down because we have no additional code load / JIT compilations. In 32-core case, we can see that peak performance suffers a bit, since preload code is not 100% efficient. The combination of these two factors is a net benefit in 2-core case: not doing JIT compilations more than pays for preload code inefficiency. This tradeoff of course depends on how well-trained the scenario is.

shipilev:
Plus, preload code needs to be a bit more resilient to invalidations, which would improve the perf gap here: https://github.com/openjdk/leyden/pull/38. With that POC patch applied, the peak performance gap shortens considerably:

```
# === 32 cores

# Default
  Time (mean ± σ):     404.0 ms ±  17.0 ms    [User: 1070.5 ms, System: 168.6 ms]
  Range (min … max):   381.0 ms … 464.8 ms    30 runs
 
# -XX:+PreloadOnly
  Time (mean ± σ):     415.1 ms ±   5.8 ms    [User: 525.3 ms, System: 82.5 ms]
  Range (min … max):   403.8 ms … 428.4 ms    30 runs

# === 2 cores

# Default
  Time (mean ± σ):     533.0 ms ±  29.3 ms    [User: 867.6 ms, System: 114.9 ms]
  Range (min … max):   463.6 ms … 587.6 ms    30 runs

# -XX:+PreloadOnly
  Time (mean ± σ):     404.5 ms ±   4.1 ms    [User: 508.8 ms, System: 78.4 ms]
  Range (min … max):   399.6 ms … 419.4 ms    30 runs
```

vnkozlov (approved):
Looks good. Next step could be to cache only preload code during training to reduce size of cached data and time to load it. You don't need to save profiling data too.

shipilev:
> Next step could be to cache only preload code during training to reduce size of cached data and time to load it. You don't need to save profiling data too.

Yes, that would a good next step if we find this thing useful.

Thanks for the review!

/integrate

iwanowww:
Proposed solution looks to me way too specific and niche. 

Instead, I'd prefer to see a way to combine different JIT-compilation modes (`-XX:TieredStopAtLevel=01234`) with ability to limit usage of AOTed code (e.g., `-XX:DisableAOTCodeLevels=P124`).  `PreloadOnly` is equivalent to `-XX:TieredStopAtLevel=0` and `-XX:DisableAOTCodeLevels=124`.

PR #45, opened 2025-03-02
8351004: [leyden] Add test cases for cached Reference objects

iklam:
[JDK-8341587](https://bugs.openjdk.org/browse/JDK-8341587) allows Soft/Weak `Reference` objects to be stored in the AOT cache. Currently Soft/Weak references are used only for supporting method handles. However, we need to make sure that the support for cached Reference is correctly implemented:

1. When we add other types of objects to the AOT cache that use `Reference` objects (e.g., [JDK-8351005](https://bugs.openjdk.org/browse/JDK-8351005) "Revert back to SoftReference for Class::reflectionData"), they should work as expected.
2. The cached `Reference` objects used by the method handles implementation (such as those used by `MethodType.internTable`) should not be unnecessarily coupled with unrelated Reference (e.g., via the `Reference::link` field due to operations in `java.lang.ref.ReferenceQueue` or `java.lang.ref.Finalizer`). Otherwise, this could cause unrelated objects to be unintentionally stored in the AOT cache.
3. `java.lang.ref.Cleaner` should work as expected during both the AOT assembly phase and production run.
4. Finalization should work as expected during both the AOT assembly phase and production run.

This RFE adds a flag (`-XX:AOTInitTestClass=...`, available **only** in debug builds) to store arbitrary heap objects into the AOT cache. With this flag, we can the behavior of `Reference` in the AOT cache so that we can determine if the current support for Reference objects in Leyden is good enough for upstreaming to the mainline.

This RFE doesn't not test everything as listed above cases. Some additional test cases may be added by a subsequent commit.

I have observed the following:

**[A]** Cached `WeakReference` objects seem to be supported by the GC in the production run. See `testWeakReferenceCollection()`: if a referent is no longer reachable, `ref.get()` returns `null`.

**[B]** Case (2) doesn't seem to be a concern: `testQueue()` shows that the assembly phase won't accidentally find unrelated `WeakReference` objects even if they share the same queue as a `WeakReference` that's destined to be cached. See comments in `testQueue()` for more details.

**[C]** `MethodType.internTable` is a `ReferencedKeySet` that internally uses `WeakReference` to automatically remove elements that are no longer referenced. However, with `grep -n referent.*null cds.oops.txt` in the test's output directory, we can see a few `WeakReferenceKeys` whose `referent` has been collected, but we didn't remove these keys from the `internTable` at `(0xfffcd3e1)`:

```
0x00000007ffe6b2e8: @@ Object  jdk.internal.util.WeakReferenceKey
 - klass: 'jdk/internal/util/WeakReferenceKey' 0x00000008002f9ad8
 - fields (4 words):
 - private 'referent' 'Ljava/lang/Object;' @12 null
 - volatile 'queue' 'Ljava/lang/ref/ReferenceQueue;' @16 0x00000007ffe6afb0 (0xfffcd5f6) java.lang.ref.ReferenceQueue
 - volatile 'next' 'Ljava/lang/ref/Reference;' @20 null
 - private transient 'discovered' 'Ljava/lang/ref/Reference;' @24 0x00000007ffe6b308 (0xfffcd661) jdk.internal.util.WeakReferenceKey
 - private final 'hashcode' 'I' @28  52198401 (0x031c7c01)

0x00000007ffe6afb0: @@ Object (0xfffcd5f6) java.lang.ref.ReferenceQueue
 - klass: 'java/lang/ref/ReferenceQueue' 0x000000080014f178
 - fields (4 words):
 - private volatile 'head' 'Ljava/lang/ref/Reference;' @12 null
 - private 'queueLength' 'J' @16  0 (0x0000000000000000)
 - private final 'lock' 'Ljava/lang/ref/ReferenceQueue$Lock;' @24 0x00000007ffee0c78 (0xfffdc18f) java.lang.ref.ReferenceQueue$Lock

0x00000007ffe69f08: @@ Object (0xfffcd3e1) jdk.internal.util.ReferencedKeyMap
 - klass: 'jdk/internal/util/ReferencedKeyMap' 0x00000008002f7298
 - fields (3 words):
 - private final 'isSoft' 'Z' @12  false (0x00)
 - private final 'map' 'Ljava/util/Map;' @16 0x00000007ffe69f20 (0xfffcd3e4) java.util.concurrent.ConcurrentHashMap
 - private final 'stale' 'Ljava/lang/ref/ReferenceQueue;' @20 0x00000007ffe6afb0 (0xfffcd5f6) java.lang.ref.ReferenceQueue
```

The `(0xfffcd65d) WeakReferenceKey` should have been added to the `ReferencedKeyMap::stale` queue, but we can see that the queue is empty.

We can see a few non-null  `Reference::discovered` field in `cds.oops.txt`. These are waiting for the `Reference$ReferenceHandler` thread to move them onto the target `ReferenceQueue.`

However, take a look at the following two lines in the `WeakReferenceTestApp.aot.log` file from the test's output directory: The `Reference$ReferenceHandler` thread is disabled in the assembly phase.

```
[0.137s][info ][cds] JVM_StartThread() ignored: java.lang.ref.Reference$ReferenceHandler
[0.137s][info ][cds] JVM_StartThread() ignored: java.lang.ref.Finalizer$FinalizerThread
```

There are two reasons that we run only a single Java thread during the assembly phase:

- Historically, we wanted the contents of the CDS archive deterministic. Concurrently executing Java threads made this very difficult, so we basically hijacked `Thread::start()` to disallow the launching of new Java threads.
- Right before we enter a safepoint to dump the AOT cache, we execute a small amount of Java code (in the main Java thread) to clean up global states. Concurrent Java threads may step on these clean up operations.

Maybe we can enable multiple Java threads for Leyden

- With new Leyden optimizations such as profiling and AOT compilation, the AOT cache is no longer deterministic anyway. It will be exceedingly difficult to get deterministic contents.
- As @shipilev mentioned offline, we may need a way for the GC to tell us that it "has finished all (concurrent) work". At that point, we can wait for the ReferenceHandler/FinalizerThread to become quiescent before we start the global clean up operations. We need to restrict these operations such that they won't cause additional GCs, etc.

Also, we need to call `jdk.internal.util.ReferencedKeyMap::removeStaleReferences()` during the global clean up:

```
    public void removeStaleReferences() {
        while (true) {
            Object key = stale.poll();
            if (key == null) {
                break;
            }
            map.remove(key);
        }
    }
```

---------
### Progress

### Issue
 * [JDK-8351004](https://bugs.openjdk.org/browse/JDK-8351004): [leyden] Add test cases for cached Reference objects (**Enhancement** - P4)

### Reviewers
 * [John R Rose](https://openjdk.org/census#jrose) (@rose00 - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/45/head:pull/45` \
`$ git checkout pull/45`

`$ git checkout pull/45` \
`$ git pull https://git.openjdk.org/leyden.git pull/45/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 45`

`$ git pr show -t 45`

</details>

<a href="https://git.openjdk.org/leyden/pull/45.diff">https://git.openjdk.org/leyden/pull/45.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/45#issuecomment-2692571210)
</details>

### DIFF
--- a/src/hotspot/share/cds/aotClassInitializer.cpp
+++ b/src/hotspot/share/cds/aotClassInitializer.cpp
@@ -27,6 +27,7 @@
 #include "cds/cdsConfig.hpp"
 #include "dumpTimeClassInfo.inline.hpp"
 #include "cds/heapShared.hpp"
+#include "classfile/symbolTable.hpp"
 #include "classfile/systemDictionaryShared.hpp"
 #include "classfile/vmSymbols.hpp"
 #include "memory/resourceArea.hpp"
@@ -34,8 +35,11 @@
 #include "oops/instanceKlass.inline.hpp"
 #include "oops/symbol.hpp"
 #include "runtime/fieldDescriptor.inline.hpp"
+#include "runtime/java.hpp"
 #include "runtime/javaCalls.hpp"
 #include "runtime/mutexLocker.hpp"
+
+DEBUG_ONLY(InstanceKlass* _aot_init_class = nullptr;)
 
 // Detector for class names we wish to handle specially.
 // It is either an exact string match or a string prefix match.
@@ -321,6 +325,12 @@ bool AOTClassInitializer::can_archive_initialized_mirror(InstanceKlass* ik) {
     }
   }
 
+#ifdef ASSERT
+  if (ik == _aot_init_class) {
+    return true;
+  }
+#endif
+
   return false;
 }
 
@@ -479,3 +489,33 @@ void AOTClassInitializer::maybe_preinit_class(InstanceKlass* ik, TRAPS) {
   }
 #endif
 }
+
+#ifdef ASSERT
+void AOTClassInitializer::init_test_class(TRAPS) {
+  // -XX:AOTInitTestClass is used in regression tests for adding additional AOT-initialized classes
+  // and heap objects into the AOT cache. The tests must be carefully written to avoid including
+  // any classes that cannot be AOT-initialized.
+  //
+  // -XX:AOTInitTestClass is NOT a general mechanism for including user-defined objects into
+  // the AOT cache. Therefore, this option is NOT available in product JVM.
+  if (AOTInitTestClass != nullptr && CDSConfig::is_initing_classes_at_dump_time()) {
+    log_info(cds)("Debug build only: force initialization of AOTInitTestClass %s", AOTInitTestClass);
+    TempNewSymbol class_name = SymbolTable::new_symbol(AOTInitTestClass);
+    Handle app_loader(THREAD, SystemDictionary::java_system_loader());
+    Klass* k = SystemDictionary::resolve_or_null(class_name, app_loader, CHECK);
+    if (k == nullptr) {
+      vm_exit_during_initialization("AOTInitTestClass not found", AOTInitTestClass);
+    }
+    if (!k->is_instance_klass()) {
+      vm_exit_during_initialization("Invalid name for AOTInitTestClass", AOTInitTestClass);
+    }
+
+    _aot_init_class = InstanceKlass::cast(k);
+    _aot_init_class->initialize(CHECK);
+  }
+}
+
+bool AOTClassInitializer::has_test_class() {
+  return _aot_init_class != nullptr;
+}
+#endif
--- a/src/hotspot/share/cds/aotClassInitializer.hpp
+++ b/src/hotspot/share/cds/aotClassInitializer.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2024, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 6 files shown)


rose00:
This is excellent work.  I agree that we must be cautious about executing arbitrary code in the assembly phase, and that includes reference queue logic.

That said, I think we can largely trust what the reference queues are doing here, and therefore benefit from their work.

The big goal here is to disallow arbitrary user-defined code in the assembly phase.  A secondary goal (which helps use learn how to meet the big goal) is keeping tight control on which JDK-defined code gets run.

Both those goals suggest that the problem is not running reference queue threads, but rather the problem is restricting what happens when a cleaner or finalizer runs.  While the assembly phase is running, I think it would be good to "watch carefully" in the reference processor code (which we trust) for attempts to execute finalizers or cleaners which are not on our "OK list" of cleaners and finalizers.  Even if JDK code defines them, if they would surprise us, the surprise needs to be conveyed in the form of an error, saying "hey, I know this is JDK logic, but you didn't list it as OK to run".

Can we install such a test in the RQ code?  Find the class responsible, and ask if it is on our short list of trusted RQ handler classes?

rose00:
That said, I think this PR can be pushed, even with the FIXMEs.  You might need to open a second tracking bug for the remaining work. That remaining work may include my suggestion, of adding error-detection logic in the reference queue processors defined by the JDK.

rose00 (approved):

iklam:
Thanks @rose00 for the review. I will try to work on the ReferenceQueue processing in a future RFE.
/integrate

PR #46, opened 2025-03-04
8347755: Support static library in jmod

slowhog:
This PR add --static-libs option to the jmod tool, it's basically mirroring --libs option to support adding a new section of static archive into the jmod file under the new section `static-lib`.

The JMOD magic header contains a MAJOR and MINOR version, we bump up the MINOR version if the --static-libs option is specified; otherwise, keep the MINOR version as 1. This allow created JMOD file without the newly supported static-lib section to continue be consumed by earlier version of jmod tool.

Also fix the formatting for _hermetic_jdk_jimage_offset, which is a julong(uint64_t) as unsigned long long on MacOS, thus need to use %llu.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8347755](https://bugs.openjdk.org/browse/JDK-8347755): Support static library in jmod (**Enhancement** - P4)

### Reviewers
 * [Ron Pressler](https://openjdk.org/census#rpressler) (@pron - Committer)
 * [Severin Gehwolf](https://openjdk.org/census#sgehwolf) (@jerboaa - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/46/head:pull/46` \
`$ git checkout pull/46`

`$ git checkout pull/46` \
`$ git pull https://git.openjdk.org/leyden.git pull/46/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 46`

`$ git pr show -t 46`

</details>

<a href="https://git.openjdk.org/leyden/pull/46.diff">https://git.openjdk.org/leyden/pull/46.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/46#issuecomment-2696055659)
</details>


test/jdk/tools/jmod/JmodTest.java
```diff
@@ -920,6 +956,20 @@ static void assertSameContent(Path p1, Path p2) {
         }
     }
 
+    static void assertMagic(Path file, int minorVersion) throws IOException {
```
AlanBateman, line 959: testLibs and testStaticLibs checking the magic number is good, just might be better to have the two call assertMagic with both the expected major and minor version rather than just the minor version.

AlanBateman (commented):
The updates to the JMOD format and jmod tool looks good, as is bumping the version when the JMOD includes a static-libs section. 

Doesn't have to be this PR but jmod.md will need to be updated if the overall feature is eventually proposed for main line.

slowhog:
> Doesn't have to be this PR but jmod.md will need to be updated if the overall feature is eventually proposed for main line.

Also CSR would be needed.

src/java.base/share/classes/jdk/internal/jmod/JmodFile.java
```diff
@@ -182,8 +187,12 @@ public JmodFile(Path file) throws IOException {
...
     }
 
-    public static void writeMagicNumber(OutputStream os) throws IOException {
-        os.write(JMOD_MAGIC_NUMBER);
+    public static void writeMagicNumber(OutputStream os, boolean withStaticLib) throws IOException {
+        if (withStaticLib) {
+            os.write(JMOD_MAGIC_NUMBER);
+        } else {
+            os.write(JMOD_MAGIC_NO_STATIC_LIB);
+        }
```
jerboaa, line 195: This conditionality seems a bit strange. Isn't the idea for the magic to be changed when the format got extended to do more things? I.e. shouldn't this be micro version `0x01` in both cases? Then when dealing with version JMOD `1.1` version you'd be checking if static-libs is there.

slowhog: The idea is without static lib, it's essential the same as the old format, using the old version so that old tools can be used if static is not included.

AlanBateman: (Just to be clear, this is about the JMOD format, not the jimage format. The jimage format is JDK internal and we are free to change it from release to release, even build to build, without compatibility concerns.)

You are right that the JMOD format is not documented but care is required when rev'ing the format because there are JDK tools that produce and consume this format. It should be possible for a project to produce a module in JMOD format with the `jmod` tool from JDK N, and put the package module on the module path when creating a run-time image with JDK N+1 `jlink`. So what Henry has looks right. It means the `jmod` tool will produce a 1.0 format when creating a package module that doesn't have static libs, and a JMOD in the new format when the module has static libs. This seems preferable to adding a `--release` option.

bowbahdoe: I guess its worth asking how would the format evolve in the future if more than `/static-lib` were added? Just as a strawman: say `/docs` and `/sources` were added in some sequence. Would this grow to say 

```
if (<nothing new>) {
   // 1.0
}
else if (<just static-lib>) {
   // 1.1
}
else {
   // 1.2
}
```

What if `/sources` (and corresponding generation of `src.zip`) were added first and `/docs` second?

```
if (<nothing new>) {
   // 1.0
}
else if (<just static-lib>) {
   // 1.1
}
else if (<just sources +/- static-lib>) {
   // 1.2
}
else {
   // 1.3
}
```

Or would, at that point, versioning move to be closer to what is done for classfiles.

AlanBateman: > I guess its worth asking how would the format evolve in the future if more than `/static-lib` were added? 

Unlike the JAR format, the JMOD format requires an update when new sections are added. We don't need to get hung up on any of the details around this now, most of the work to support this will be in jlink, not the jmod tool.

jerboaa: I'm ok with this for the leyden branch.

slowhog: I am thinking just to support -1 version. So 1.2 jmod can produce 1.2 with latest feature and 1.1 for previous version.
Anyway, as Alan pointed out, we don't need to figure this out right now.

slowhog:
Latest commit to support creating of static-jmods with `make static-jmods`, thanks @magicus for help.
Result on MacOS,

```
ls -l build/macosx-x86_64-server-release/images/static-jmods   
total 184152
-rw-r--r--  1 hjen  staff  21778174 Mar 13 10:35 java.base.jmod
-rw-r--r--  1 hjen  staff    140500 Mar 13 10:35 java.compiler.jmod
-rw-r--r--  1 hjen  staff     58934 Mar 13 10:35 java.datatransfer.jmod
-rw-r--r--  1 hjen  staff  16688299 Mar 13 10:35 java.desktop.jmod
-rw-r--r--  1 hjen  staff     68824 Mar 13 10:35 java.instrument.jmod
...
-rw-r--r--  1 hjen  staff     49946 Mar 13 10:35 jdk.xml.dom.jmod
-rw-r--r--  1 hjen  staff    112583 Mar 13 10:35 jdk.zipfs.jmod
```


```
./build/macosx-x86_64-server-release/jdk/bin/jmod list build/macosx-x86_64-server-release/images/static-jmods/java.base.jmod | grep static-lib
static-lib/libjava.a
static-lib/libjimage.a
static-lib/libjli.a
static-lib/libjsig.a
static-lib/libnet.a
static-lib/libnio.a
static-lib/libosxsecurity.a
static-lib/libsyslookup.a
static-lib/libverify.a
static-lib/libzip.a
static-lib/module-included-libs.txt
```

jianglizhou:
> ./build/macosx-x86_64-server-release/jdk/bin/jmod list build/macosx-x86_64-server-release/images/static-jmods/java.base.jmod | grep static-lib
> static-lib/libjava.a
> static-lib/libjimage.a
> static-lib/libjli.a
> static-lib/libjsig.a
> static-lib/libnet.a
> static-lib/libnio.a
> static-lib/libosxsecurity.a
> static-lib/libsyslookup.a
> static-lib/libverify.a
> static-lib/libzip.a
> static-lib/module-included-libs.txt

I think `libjvm.a` should be included in java.base.jmod as well.

slowhog:
> I think `libjvm.a` should be included in java.base.jmod as well.

You are right. Looks libjvm.a is not build as a dependency, and I didn't try to copy that file either.
Looks to me, it is only built for static launcher, I'll try to add that for java.base-static-jmod.

slowhog:
Verified with `make java.base-static-jmod`, and libjvm.a is included.

```
./build/macosx-x86_64-server-release/jdk/bin/jmod list build/macosx-x86_64-server-release/images/static-jmods/java.base.jmod | grep static-lib
static-lib/libjava.a
static-lib/libjimage.a
static-lib/libjli.a
static-lib/libjsig.a
static-lib/libjvm.a
static-lib/libnet.a
static-lib/libnio.a
static-lib/libosxsecurity.a
static-lib/libsyslookup.a
static-lib/libverify.a
static-lib/libzip.a
static-lib/module-included-libs.txt
```

magicus:
`static-lib/module-included-libs.txt` should not be included. It is a built-time support file. I guess this happens since you re-used an existing directory structure that was only used for build support files before. The proper fix is to make sure this file ends up in a different directory.

slowhog:
> `static-lib/module-included-libs.txt` should not be included. It is a built-time support file. I guess this happens since you re-used an existing directory structure that was only used for build support files before. The proper fix is to make sure this file ends up in a different directory.

I didn't particularly filter-out or move this file because I wonder if this file serve as a manifest. Otherwise, we could do something like how we process other folders like modules_libs. `jmod --exclude`  could be a solution as well.

make/CreateJmods.gmk
```diff
@@ -77,6 +85,36 @@ ifneq ($(LIBS_DIR), )
...
+        FLATTEN := true, \
+        SRC := $(SUPPORT_OUTPUTDIR)/native/$(MODULE), \
+        DEST := $(SUPPORT_OUTPUTDIR)/modules_static-libs/$(MODULE), \
+        FILES := $(STATIC_LIB_FILES), \
+    ))
+    DEPS += $(COPY_STATIC_LIBS)
+    JMOD_FLAGS += --static-libs $(STATIC_LIBS_DIR)
+  endif
+
+  # Copy hotspot libs, currently only libjvm.a
```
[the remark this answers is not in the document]
magicus: Should we really, though? That includes a `main` function. If the user wants to use the static libs from the jmod to create their own application, that will cause a symbol conflict.

magicus:
> I didn't particularly filter-out or move this file because I wonder if this file serve as a manifest. 

No, it doesn't, it's just an implementation specific detail of the build system.

> Otherwise, we could do something like how we process other folders like modules_libs. `jmod --exclude` could be a solution as well.

No, no. We should just don't mix temporary support files with the proper output files.

make/CreateJmods.gmk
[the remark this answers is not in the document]
slowhog: I think that worth a discussion, IIRC, the most needed processing should be in libjli already.

magicus:
However, regarding manifests, there *is* one thing that really should be included with the static libs, and that is a `pkg-config` file, that tells the end users what linker flags to use to be able to properly use the static libraries. 

I have been planning to add this to the static-jdk-image but never gotten around to it. (Can't find the JBS number now either.) But I guess that there is no real conflict between that and this PR; if I just plop the file down alongside the .a files, it will be automatically picked up by jmod.

slowhog:
> > I didn't particularly filter-out or move this file because I wonder if this file serve as a manifest.
> 
> No, it doesn't, it's just an implementation specific detail of the build system.
> 

I know the original intention of this is not for that. What I am asking is if we should have such a manifest or hints for user? Similar to pkg-config.

> > Otherwise, we could do something like how we process other folders like modules_libs. `jmod --exclude` could be a solution as well.
> 
> No, no. We should just don't mix temporary support files with the proper output files.

That I agree.

magicus:
> Similar to pkg-config.

Maybe we just wrote past each other here... As I said, my plan is to actually include a pkg-config file. This is a canonical way to describe exactly what is needed to utilize a library. There is no reason to do "something almost like pkg-config but not actually pkg-config". :-)

make/CreateJmods.gmk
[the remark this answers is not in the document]
magicus: Yes, main.c is basically a thin wrapper that parses command line arguments and calls `JLI_Launch`.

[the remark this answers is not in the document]
magicus: Sorry @jianglizhou, I did not properly read the second part of your suggestion, to create a `liblauncher.o` out of `main.o`. Well, tbh, this is just JDK-8351367 all over again. If we should include "optional" libraries in the jmods, we need to figure out first how to handle them. How should they be marked/classified as optional or required? How should the user tell jlink to include or exclude them? If we do get a good answer to that question, then we can introduce an optional `liblauncher.a` as well alongside `libjsig.a`. But until then, we should just stay clear of "optional" static libraries.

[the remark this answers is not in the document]
jianglizhou: I think for common cases, there would be no C/C++ `main()` function provided from user code with Java applications. The JDK `java` launcher's `main()` is the C/C++ entry point. So we do need to provide a static library with the C/C++ `main` function for linking the executable.

With the customized launcher cases, the C/C++ `main()` is implemented by the customized launcher. Then, the JDK provided `main` entry point is not needed.

@magicus, yes, this involves with the optional static library in the jmods, as you pointed out. I think for C/C++ `main`, supporting that is necessary.

An alternative is to require Java application implement C/C++ `main()`. That doesn't seem to be an applaudable approach.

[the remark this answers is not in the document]
slowhog: I would prefer that thin wrapper is to be provided by the end tools(like jlink) which use the jmod to link an executable.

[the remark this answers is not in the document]
jianglizhou: > I would prefer that thin wrapper is to be provided by the end tools(like jlink) which use the jmod to link an executable.

@slowhog Can you provide details on that? Do you suggest there would be a separate executable that provides the C/C++ `main` entry point as a thin wrapper?  The binary that statically linked with the `libjvm.a`, `libjli.a` and etc would contain no C/C++ `main`? Is that binary linked as a shred library? Or, your proposal is to generate some C/C++ code with the `main()` function using jlink, then compile the C/C++ code and use that to link the executable?

The issue that we want to solve here is to provide the C/C++ `main()` entry point for the executable for the final image.

jianglizhou:
> @slowhog This pull request has been inactive for more than 4 weeks and will be automatically closed if another 4 weeks passes without any activity. To avoid this, simply add a new comment to the pull request. Feel free to ask for assistance if you need help with progressing this pull request towards integration!

@slowhog Could you please add a comment to make sure your PR is not closed? Thanks

slowhog:
We agreed that to add a liblauncher to facilitate the `main` entry point. However, as I mentioned before, this is really a depends on situation and should be answered at link time. One obvious question is how do we deliver javaw(with WinMain) symbol for Windows?

Also all the tunable configuration at build time such as MAIN_CLASS and JAVA_ARGS for an executable? Or are we simply thinking of `java` here?

jianglizhou:
> We agreed that to add a liblauncher to facilitate the `main` entry point. However, as I mentioned before, this is really a depends on situation and should be answered at link time. One obvious question is how do we deliver javaw(with WinMain) symbol for Windows?
> 
> Also all the tunable configuration at build time such as MAIN_CLASS and JAVA_ARGS for an executable? Or are we simply thinking of `java` here?

The default Java launcher `main.o` (and packaged as `liblauncher` static library) is produced by https://github.com/openjdk/jdk/blob/ad7c475fb1e23f583a33d58f0bd73ea0fb56740c/make/modules/java.base/Launcher.gmk#L39:

```
$(eval $(call SetupBuildLauncher, java, \
    ENABLE_ARG_FILES := true, \
    EXPAND_CLASSPATH_WILDCARDS := true, \
    EXTRA_RCFLAGS := $(JAVA_RCFLAGS), \
    VERSION_INFO_RESOURCE := $(JAVA_VERSION_INFO_RESOURCE), \
    OPTIMIZATION := HIGH, \
))
```

That should be used to link the launcher executable when jlink creates the final image, if no custom launcher is used.

With the https://github.com/openjdk/leyden/blob/hermetic-java-runtime/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/LinkHermeticImagePlugin.java prototype, we could pass additional linking flags to to form the complete native linking command (when integrating with the static .jmod support). With with that, I think we can pass the default `liblauncher.a` as part of the extra linking flags. Then, that may help simplify things for us by not having to include `liblauncher` in `java.base` static .jmod and not have to deal with optionally using the default `liblauncher` from the .jmod for linking. We could go with that. 

The default launcher and custom launcher would be handled similarly then, by passing via the extra linking flag.

slowhog:
> The default Java launcher `main.o` (and packaged as `liblauncher` static library) is produced by https://github.com/openjdk/jdk/blob/ad7c475fb1e23f583a33d58f0bd73ea0fb56740c/make/modules/java.base/Launcher.gmk#L39:
> 
> ```
> $(eval $(call SetupBuildLauncher, java, \
>     ENABLE_ARG_FILES := true, \
>     EXPAND_CLASSPATH_WILDCARDS := true, \
>     EXTRA_RCFLAGS := $(JAVA_RCFLAGS), \
>     VERSION_INFO_RESOURCE := $(JAVA_VERSION_INFO_RESOURCE), \
>     OPTIMIZATION := HIGH, \
> ))
> ```
> 
> That should be used to link the launcher executable when jlink creates the final image, if no custom launcher is used.
> 
> With the https://github.com/openjdk/leyden/blob/hermetic-java-runtime/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/LinkHermeticImagePlugin.java prototype, we could pass additional linking flags to to form the complete native linking command (when integrating with the static .jmod support). With with that, I think we can pass the default `liblauncher.a` as part of the extra linking flags. Then, that may help simplify things for us by not having to include `liblauncher` in `java.base` static .jmod and not have to deal with optionally using the default `liblauncher` from the .jmod for linking. We could go with that.
> 
> The default launcher and custom launcher would be handled similarly then, by passing via the extra linking flag.

Yes, that's just for the `java` launcher. If that's the only executable for java.base static mod targeted for. `javaw` would be another executable in java.base, so we will need another one for that.
Regard to other executable, this approach would expect they would need to build a separate lib in their own corresponding static jmod to link with.

jianglizhou:
> > The default Java launcher `main.o` (and packaged as `liblauncher` static library) is produced by https://github.com/openjdk/jdk/blob/ad7c475fb1e23f583a33d58f0bd73ea0fb56740c/make/modules/java.base/Launcher.gmk#L39:
> > ```
> > $(eval $(call SetupBuildLauncher, java, \
> >     ENABLE_ARG_FILES := true, \
> >     EXPAND_CLASSPATH_WILDCARDS := true, \
> >     EXTRA_RCFLAGS := $(JAVA_RCFLAGS), \
> >     VERSION_INFO_RESOURCE := $(JAVA_VERSION_INFO_RESOURCE), \
> >     OPTIMIZATION := HIGH, \
> > ))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That should be used to link the launcher executable when jlink creates the final image, if no custom launcher is used.
> > With the https://github.com/openjdk/leyden/blob/hermetic-java-runtime/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/LinkHermeticImagePlugin.java prototype, we could pass additional linking flags to to form the complete native linking command (when integrating with the static .jmod support). With with that, I think we can pass the default `liblauncher.a` as part of the extra linking flags. Then, that may help simplify things for us by not having to include `liblauncher` in `java.base` static .jmod and not have to deal with optionally using the default `liblauncher` from the .jmod for linking. We could go with that.
> > The default launcher and custom launcher would be handled similarly then, by passing via the extra linking flag.
> 
> Yes, that's just for the `java` launcher. If that's the only executable for java.base static mod targeted for. `javaw` would be another executable in java.base, so we will need another one for that. 

`javaw` is Windows only, right? It would only need to be included for Windows build. 

> Regard to other executable, this approach would expect they would need to build a separate lib in their own corresponding static jmod to link with.

Yes. Or they could just directly pass the user supplied static libraries or .o files to jlink via the extra linking flag.

pron (approved):

jerboaa (approved):

jianglizhou:
I'm looking to using the static libs from the static jmods for linking launcher executable. I just pushed https://github.com/openjdk/leyden/commit/664d3070dcc8040dacbb2adc20245e9cd580fe65 with following changes for static jmods. Do we want to start review process for all changes to the leyden/hermetic-java-runtime branch now?

- Build static-jmods as a dependency for static-jdk.
- Copy the static jmods into static-jdk/jmods, instead of copying from the regular JDK.

The shared libraries (`.so`) are still included in the static jmods. I think only the static libraries (`.a`) are needed in the static jmods.
```
$ jmod list ./linux-x86_64-server-fastdebug/images/static-jdk/jmods/java.base.jmod | grep libjvm
lib/server/libjvm.so
static-lib/libjvm.a
```

PR #47, opened 2025-03-06
8351493: Revert build changes that are no longer needed

magicus:
I compared the hermetic-java-runtime branch with mainline, and noticed some old build changes lingering, that are no longer needed. Furthermore, I checked a handful of other files and noticed a few spurious changes that just make the branch differ from mainline for no good reason. I also reverted those.

With this patch, the difference between hermetic-java-runtime and mainline gets even more minimal.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8351493](https://bugs.openjdk.org/browse/JDK-8351493): Revert build changes that are no longer needed (**Bug** - P4)

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - **Reviewer**)
 * [Jiangli Zhou](https://openjdk.org/census#jiangli) (@jianglizhou - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/47/head:pull/47` \
`$ git checkout pull/47`

`$ git checkout pull/47` \
`$ git pull https://git.openjdk.org/leyden.git pull/47/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 47`

`$ git pr show -t 47`

</details>

<a href="https://git.openjdk.org/leyden/pull/47.diff">https://git.openjdk.org/leyden/pull/47.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/47#issuecomment-2710532121)
</details>


magicus:
There were a lot of build failures on GHA, but that seem to be the cause for the hermetic-java-runtime branch even without this patch.

magicus:
@jianglizhou I'd like your review as well, please?

jianglizhou:
@magicus Apologize for the delay, as I didn't realize that you were looking for review on the change. The removal looks fine to me. I initially added `--with-static-java` to explicitly build static libraries that could be handled as system provided instead of bundled by default. I haven't removed it as I was still considering (in background) if it would be needed. We can always add a new flag when we identify the usages.

jianglizhou (approved):

magicus:
@jianglizhou 
> [...] explicitly build static libraries that could be handled as system provided instead of bundled by default.

I'm not sure what you mean by this?

magicus:
I don't really see how this differs from the dynamic case. The distributor building the JDK must decide whether to use the system libraries (typically done by Linux distributions, since they have control over such things), or bundle it (most other). The same goes for when building a static library. Even if we allow some kind of "mix and match" with jlink, there needs to be some kind of limit on how much a single version of the JDK can provide for. In one extreme, mixing different address sizes or CPU architectures is not possible, and will never be. On the other hand, libjsig has been provided as an optional library. So somewhere between there will libharfbuzz need to fall. My instinct is that it is not like libjsig but rather like the OS; that is, for a specific instance of the JDK a distributor builds, they decide not only OS and CPU, but also if libraries like harfbuzz should be included or not, both for static and dynamic.

jianglizhou:
> I don't really see how this differs from the dynamic case. The distributor building the JDK must decide whether to use the system libraries (typically done by Linux distributions, since they have control over such things), or bundle it (most other). The same goes for when building a static library. Even if we allow some kind of "mix and match" with jlink, there needs to be some kind of limit on how much a single version of the JDK can provide for. In one extreme, mixing different address sizes or CPU architectures is not possible, and will never be. On the other hand, libjsig has been provided as an optional library. So somewhere between there will libharfbuzz need to fall. My instinct is that it is not like libjsig but rather like the OS; that is, for a specific instance of the JDK a distributor builds, they decide not only OS and CPU, but also if libraries like harfbuzz should be included or not, both for static and dynamic.

Thanks for the reasoning with dynamic case. It does make sense to treat libharfbuzz and other bundled libs on static the same as on dynamic JDK.

PR #48, opened 2025-03-13
8351965: [leyden] Skip C2 AOT code if C2 preload AOT code trapped

shipilev:
This is clearly visible in compilation logs:

```
     43    W0.1    Q8.1    C0.0    293      AP 4       com.sun.tools.javac.util.StringNameTable::fromString (50 bytes)
[0.049s][debug  ][deoptimization] cid= 293     level=4 com.sun.tools.javac.util.StringNameTable::fromString(Ljava/lang/String;)Lcom/sun/tools/javac/util/Name; trap_bci=28 unloaded reinterpret pc=0x00007c6bd7e4e7ac relative_pc=0x000000000000068c
     49                            293      AP 4       com.sun.tools.javac.util.StringNameTable::fromString (50 bytes)   made not entrant
     49    W0.2    Q0.0    C0.3   1394         2       com.sun.tools.javac.util.StringNameTable::fromString (50 bytes)
     90                           1394         2       com.sun.tools.javac.util.StringNameTable::fromString (50 bytes)   made not entrant
     90    W0.0    Q0.0    C0.1   1867      A  4       com.sun.tools.javac.util.StringNameTable::fromString (50 bytes)
[0.098s][debug  ][deoptimization] cid=1867     level=4 com.sun.tools.javac.util.StringNameTable::fromString(Ljava/lang/String;)Lcom/sun/tools/javac/util/Name; trap_bci=28 unloaded reinterpret pc=0x00007c6bd7ebcb58 relative_pc=0x00000000000005d8
     98                           1867      A  4       com.sun.tools.javac.util.StringNameTable::fromString (50 bytes)   made not entrant
    104    W0.0    Q0.0    C0.3   1942         2       com.sun.tools.javac.util.StringNameTable::fromString (50 bytes)
    130                           1942         2       com.sun.tools.javac.util.StringNameTable::fromString (50 bytes)   made not entrant
    130    W1.0    Q0.7   C13.3   1968         4       com.sun.tools.javac.util.StringNameTable::fromString (50 bytes)
```

So the AP4 method was preloaded, then it trapped and got replaced by T2 method, which eventually got to C2, at which point we loaded A4 method. That method trapped _at the same bci_, so we are back at T2, then then to real T4. So we have spent one deopt cycle unnecessarily, and the code was in T2 for twice as long.

I don't think we would be able to fully tame uncommon traps from the preload code, so fixing this gap is valuable.

`decompile_count()` is only updated by C2, so we can just check it directly.

Additional testing:
 - [x] Ad-hoc benchmarks

---------
### Progress

### Issue
 * [JDK-8351965](https://bugs.openjdk.org/browse/JDK-8351965): [leyden] Skip C2 AOT code if C2 preload AOT code trapped (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/48/head:pull/48` \
`$ git checkout pull/48`

`$ git checkout pull/48` \
`$ git pull https://git.openjdk.org/leyden.git pull/48/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 48`

`$ git pr show -t 48`

</details>

<a href="https://git.openjdk.org/leyden/pull/48.diff">https://git.openjdk.org/leyden/pull/48.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/48#issuecomment-2721962065)
</details>

### DIFF
--- a/src/hotspot/share/ci/ciEnv.cpp
+++ b/src/hotspot/share/ci/ciEnv.cpp
@@ -1080,6 +1080,9 @@ void ciEnv::register_method(ciMethod* target,
       MethodData* mdo = method()->method_data();
       if (mdo != nullptr && _inc_decompile_count_on_failure) {
         mdo->inc_decompile_count();
+        if (nm->preloaded()) {
+          mdo->inc_preload_decompile_count();
+        }
       }
 
       // All buffers in the CodeBuffer are allocated in the CodeCache.
--- a/src/hotspot/share/ci/ciEnv.cpp
+++ b/src/hotspot/share/ci/ciEnv.cpp
@@ -1080,7 +1080,7 @@ void ciEnv::register_method(ciMethod* target,
       MethodData* mdo = method()->method_data();
       if (mdo != nullptr && _inc_decompile_count_on_failure) {
         mdo->inc_decompile_count();
-        if (nm->preloaded()) {
+        if (nm != nullptr && nm->preloaded()) {
           mdo->inc_preload_decompile_count();
         }
       }
--- a/src/hotspot/share/ci/ciEnv.cpp
+++ b/src/hotspot/share/ci/ciEnv.cpp
@@ -1080,9 +1080,6 @@ void ciEnv::register_method(ciMethod* target,
       MethodData* mdo = method()->method_data();
       if (mdo != nullptr && _inc_decompile_count_on_failure) {
         mdo->inc_decompile_count();
-        if (nm != nullptr && nm->preloaded()) {
-          mdo->inc_preload_decompile_count();
-        }
       }
 
       // All buffers in the CodeBuffer are allocated in the CodeCache.
--- a/src/hotspot/share/code/SCCache.cpp
+++ b/src/hotspot/share/code/SCCache.cpp
@@ -4456,3 +4456,22 @@ SCCache::ReadingMark::~ReadingMark() {
     }
   }
 }
+
+void SCCache::block_loading(Method* method, CompLevel comp_level) {
+  SCCache* cache = open_for_read();
+  if (cache == nullptr) {
+    return;
+  }
+
+  ReadingMark rdmk;
+  if (rdmk.failed()) {
+    // Cache is closed, cannot touch anything.
+    return;
+  }
+
+  Thread* current = Thread::current();
+  HandleMark hm(current);
+  methodHandle mh(current, method);
+  SCCEntry* scc_entry = cache->find_code_entry(mh, comp_level);
+  cache->invalidate_entry(scc_entry);
+}
--- a/src/hotspot/share/code/SCCache.cpp
+++ b/src/hotspot/share/code/SCCache.cpp
@@ -321,7 +321,7 @@ void SCCache::preload_code(JavaThread* thread) {
   _cache->preload_startup_code(thread);
 }
 
-SCCEntry* SCCache::find_code_entry(const methodHandle& method, uint comp_level) {
+SCCEntry* SCCache::find_code_entry(const methodHandle& method, uint comp_level, bool maybe_ignore) {
   switch (comp_level) {
     case CompLevel_simple:
       if ((DisableCachedCode & (1 << 0)) != 0) {
@@ -363,14 +363,16 @@ SCCEntry* SCCache::find_code_entry(const methodHandle& method, uint comp_level) 
 #endif
     }
 
-    DirectiveSet* directives = DirectivesStack::getMatchingDirective(method, nullptr);
-    if (directives->IgnorePrecompiledOption) {
-      LogStreamHandle(Info, scc, compilation) log;
-      if (log.is_enabled()) {
-        log.print("Ignore cached code entry on level %d for ", comp_level);
-        method->print_value_on(&log);
+    if (maybe_ignore) {
+      DirectiveSet* directives = DirectivesStack::getMatchingDirective(method, nullptr);
+      if (directives->IgnorePrecompiledOption) {
+        LogStreamHandle(Info, scc, compilation) log;
+        if (log.is_enabled()) {
+          log.print("Ignore cached code entry on level %d for ", comp_level);
+          method->print_value_on(&log);
+        }
+        return nullptr;
       }
-      return nullptr;
     }
 
     return entry;
@@ -4472,6 +4474,6 @@ void SCCache::block_loading(Method* method, CompLevel comp_level) {
   Thread* current = Thread::current();
   HandleMark hm(current);
   methodHandle mh(current, method);
-  SCCEntry* scc_entry = cache->find_code_entry(mh, comp_level);
+  SCCEntry* scc_entry = cache->find_code_entry(mh, comp_level, /* maybe_ignore = */ false);
   cache->invalidate_entry(scc_entry);
 }
--- a/src/hotspot/share/ci/ciMethodData.hpp
+++ b/src/hotspot/share/ci/ciMethodData.hpp
@@ -537,9 +537,6 @@ public:
   uint decompile_count() const {
     return _orig.decompile_count();
   }
-  uint preload_decompile_count() const {
-    return _orig.preload_decompile_count();
-  }
   uint trap_count(int reason) const {
     return _orig.trap_count(reason);
   }
--- a/src/hotspot/share/ci/ciMethodData.hpp
+++ b/src/hotspot/share/ci/ciMethodData.hpp
@@ -537,6 +537,9 @@ public:
   uint decompile_count() const {
     return _orig.decompile_count();
   }
+  uint preload_decompile_count() const {
+    return _orig.preload_decompile_count();
+  }
   uint trap_count(int reason) const {
     return _orig.trap_count(reason);
   }
--- a/src/hotspot/share/code/SCCache.hpp
+++ b/src/hotspot/share/code/SCCache.hpp
@@ -552,6 +552,8 @@ public:
 
   static bool load_nmethod(ciEnv* env, ciMethod* target, int entry_bci, AbstractCompiler* compiler, CompLevel comp_level) NOT_CDS_RETURN_(false);
 
+  static void block_loading(Method* method, CompLevel comp_level);
+
   static SCCEntry* store_nmethod(const methodHandle& method,
                      int compile_id,
                      int entry_bci,
--- a/src/hotspot/share/code/SCCache.hpp
+++ b/src/hotspot/share/code/SCCache.hpp
@@ -608,7 +608,7 @@ public:
   static bool allow_const_field(ciConstant& value) NOT_CDS_RETURN_(false);
   static void invalidate(SCCEntry* entry) NOT_CDS_RETURN;
   static bool is_loaded(SCCEntry* entry);
-  static SCCEntry* find_code_entry(const methodHandle& method, uint comp_level);
+  static SCCEntry* find_code_entry(const methodHandle& method, uint comp_level, bool maybe_ignore = true);
   static void preload_code(JavaThread* thread);
 
   template<typename Function>
--- a/src/hotspot/share/code/nmethod.cpp
+++ b/src/hotspot/share/code/nmethod.cpp
@@ -1967,9 +1967,6 @@ void nmethod::inc_decompile_count() {
   if (mdo == nullptr)  return;
   // There is a benign race here.  See comments in methodData.hpp.
   mdo->inc_decompile_count();
-  if (preloaded()) {
-    mdo->inc_preload_decompile_count();
-  }
 }
 
 void nmethod::inc_method_profiling_count() {
@@ -2076,6 +2073,15 @@ bool nmethod::make_not_entrant(const char* reason, bool make_not_entrant) {
     if (update_recompile_counts()) {
       // Mark the method as decompiled.
       inc_decompile_count();
+    }
+
+    if (preloaded() && method() != nullptr) {
+      // Preload code was decompiled, possibly due to a trap. Preload code should not
+      // normally have traps, but some are still there. Since preload code is compiled
+      // more conservatively, we assume that non-preload code would trap again in similar
+      // conditions. Therefore, there is no point in trying to load it. Instead, wait for
+      // normal compilation to occur.
+      SCCache::block_loading(method(), CompLevel_full_optimization);
     }
 
     BarrierSetNMethod* bs_nm = BarrierSet::barrier_set()->barrier_set_nmethod();
--- a/src/hotspot/share/code/nmethod.cpp
+++ b/src/hotspot/share/code/nmethod.cpp
@@ -2075,6 +2075,7 @@ bool nmethod::make_not_entrant(const char* reason, bool make_not_entrant) {
       inc_decompile_count();
     }
 
+#if INCLUDE_CDS
     if (preloaded() && method() != nullptr) {
       // Preload code was decompiled, possibly due to a trap. Preload code should not
       // normally have traps, but some are still there. Since preload code is compiled
@@ -2083,6 +2084,7 @@ bool nmethod::make_not_entrant(const char* reason, bool make_not_entrant) {
       // normal compilation to occur.
       SCCache::block_loading(method(), CompLevel_full_optimization);
     }
+#endif
 
     BarrierSetNMethod* bs_nm = BarrierSet::barrier_set()->barrier_set_nmethod();
     if (bs_nm == nullptr || !bs_nm->supports_entry_barrier(this)) {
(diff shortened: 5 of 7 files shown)


shipilev:
Some benchmark data. Not the improvement in `User` time.

`JavacBenchApp 1`

```
Benchmark 1: build/linux-x86_64-server-release/images/jdk/bin/java -Xms64m -Xmx8g -XX:+UseSerialGC -cp JavacBenchApp.jar -XX:AOTCache=app.aot JavacBenchApp 1

# Before
  Time (mean ± σ):     185.0 ms ±   2.8 ms    [User: 413.9 ms, System: 71.5 ms]
  Range (min … max):   179.8 ms … 191.2 ms    30 runs
 
# After
  Time (mean ± σ):     182.0 ms ±   2.1 ms    [User: 352.7 ms, System: 69.5 ms]
  Range (min … max):   176.7 ms … 185.8 ms    30 runs
```

vnkozlov:
> That method trapped _at the same bci_, which is really expected

Why it is expected? Going back to Interpreter or C1 code should load class and next AOT code should not trap.

shipilev:
> > That method trapped _at the same bci_, which is really expected
> 
> Why it is expected? Going back to Interpreter or C1 code should load class and next AOT code should not trap.

This is what I thought when digging through it: if there is a free-standing uncommon trap without clinit barrier on rare branch, maybe we would see the same uncommon trap from both AOT versions, whether preloaded or not preloaded. Let me dump the assembly and see what is actually going on.

iwanowww (commented):
I don't know whether it is a good heuristic for a general case. How much does it improve the behavior?

Most likely, repeated deoptimizations happen because compiled code happens to arrive much earlier than it happened during training run. (Unless training run suffered from the very same deoptimizations.) So, IMO the question boils down to whether it is beneficial to unconditionally reprofile/recompile or continue using AOTed code hoping for the best.

Also, there was a deoptimization which may already let interpreter fix some of the trap causes. 

Another point: I believe MDO should include archived data, so decompile count may report higher number than actually observed during current run.

iwanowww:
Also, don't we try to avoid uncommon traps in preloaded code these days? So, a trap in preloaded code means there's something C2 can't get rid of.

shipilev:
> Also, don't we try to avoid uncommon traps in preloaded code these days? So, a trap in preloaded code means there's something C2 can't get rid of.

I look at it separately, and there are some traps in `ciTypeFlow`, for example, which I don't yet know how to disable in preload compilations. I don't think it is impossible -- probably throwing in more clinit barriers there -- but it would take time.

shipilev:
> I don't know whether it is a good heuristic for a general case. How much does it improve the behavior?

*A lot*: https://github.com/openjdk/leyden/pull/48#issuecomment-2721949661

But I am following up on @vnkozlov's question, since it does not make sense to me why we are trapping the second time...

shipilev:
OK, look through this.

Java code:

```
    @Override
    public Name fromString(String string) {
        Name name = nameMap.get(string);
        if (name == null) {
            if (intern) {
                string = string.intern();
            }
            name = new NameImpl(this, string);  // <--- TRAPS HERE
            nameMap.put(string, name);
        }
        return name;
    }
```


Relevant bytecode:

```
  public com.sun.tools.javac.util.Name fromString(java.lang.String);
    descriptor: (Ljava/lang/String;)Lcom/sun/tools/javac/util/Name;
    flags: (0x0001) ACC_PUBLIC
    Code:
      stack=4, locals=3, args_size=2
         0: aload_0
         1: getfield      #21                 // Field nameMap:Ljava/util/HashMap;
         4: aload_1
         5: invokevirtual #29                 // Method java/util/HashMap.get:(Ljava/lang/Object;)Ljava/lang/Object;
         8: checkcast     #33                 // class com/sun/tools/javac/util/Name
        11: astore_2
        12: aload_2
        13: ifnonnull     48
        16: aload_0
        17: getfield      #25                 // Field intern:Z
        20: ifeq          28
        23: aload_1
        24: invokevirtual #35                 // Method java/lang/String.intern:()Ljava/lang/String;
        27: astore_1
        28: new           #40                 // class com/sun/tools/javac/util/StringNameTable$NameImpl <---- TRAPS HERE
        31: dup
        32: aload_0
        33: aload_1
        34: invokespecial #42                 // Method com/sun/tools/javac/util/StringNameTable$NameImpl."<init>":(Lcom/sun/tools/javac/util/StringNameTable;Ljava/lang/String;)V
```

AP 4 code:

```
  # Traps later here:
    26	     55    W0.0   Q20.7    C0.0    379 [0.056s][debug  ][deoptimization] cid= 291     level=4 com.sun.tools.javac.util.StringNameTable::fromString(Ljava/lang/String;)Lcom/sun/tools/javac/util/Name; trap_bci=28 unloaded reinterpret pc=0x00007037e802ff10 relative_pc=0x0000000000000670

Compiled method (c2)      43                            291      AP 4       com.sun.tools.javac.util.StringNameTable::fromString (50 bytes)
 ...
  # this:     rsi:rsi   = 'com/sun/tools/javac/util/StringNameTable'
  # parm0:    rdx:rdx   = 'java/lang/String'   

[Entry Point]
  # {method} {0x00000008006848b0} 'fromString' '(Ljava/lang/String;)Lcom/sun/tools/javac/util/Name;' in 'com/sun/tools/javac/util/StringNameTable'
  # this:     rsi:rsi   = 'com/sun/tools/javac/util/StringNameTable'
  # parm0:    rdx:rdx   = 'java/lang/String'   
  #           [sp+0x50]  (sp of caller)    
  0x00007037e802f8a2:   mov    0x8(%rsi),%r10d 
  0x00007037e802f8a6:   cmp    0x8(%rax),%r10d 
  0x00007037e802f8aa:   jne    0x00007037e7851d40           ;   {runtime_call Shared Runtime ic_miss_blob}
[Verified Entry Point]
  0x00007037e802f8b0:   mov    %eax,-0x14000(%rsp)
  0x00007037e802f8b7:   push   %rbp
  0x00007037e802f8b8:   sub    $0x40,%rsp  
  0x00007037e802f8bc:   cmpl   $0x0,0x20(%r15) 
  0x00007037e802f8c4:   jne    0x00007037e80301fe           ; method entry is over
  0x00007037e802f8ca:   mov    %rsi,%r14                    ; %rsi, %r14 is now "this"
  0x00007037e802f8cd:   mov    0x14(%rsi),%r10d             ; %r10 is "this.nameMap"
  0x00007037e802f8d1:   mov    0x8(%r10),%r8d               ; type-check %r10 is java/util/HashMap
  0x00007037e802f8d5:   cmp    $0x2ff6c8,%r8d               ;   {metadata('java/util/HashMap')}
  0x00007037e802f8dc:   jne    0x00007037e802fe8a
  0x00007037e802f8e2:   mov    0x24(%r10),%r11d             ; inlined HashMap.get: 
  0x00007037e802f8e6:   test   %r11d,%r11d                  ; HashMap.table == null?
  0x00007037e802f8e9:   je     0x00007037e802fdd6           ; YES, so nameMap.get returns null, jump to slow path
 ...

  0x00007037e802fdd6:   mov    %rdx,%r10            
  0x00007037e802fdd9:   movzbl 0x10(%r14),%r11d             ; %r14 is still "this", checking "this.intern"
  0x00007037e802fdde:   test   %r11d,%r11d                  ; (intern == false)?
  0x00007037e802fde1:   je     0x00007037e802fefc           ; YES, go uncommon trap
 ...
  0x00007037e802fefc:   mov    $0x28,%esi   
  0x00007037e802ff01:   mov    %r14,%rbp    
  0x00007037e802ff04:   mov    %r10,(%rsp)     
  0x00007037e802ff08:   vzeroupper 
  0x00007037e802ff0b:   call   0x00007037e7823dc0           ; ImmutableOopMap {rbp=Oop [0]=Oop } 
                                                            ;*new {reexecute=0 rethrow=0 return_oop=0}
                                                            ; - com.sun.tools.javac.util.StringNameTable::fromString@28 (line 70)
                                                            ;   {runtime_call UncommonTrapBlob}
  0x00007037e802ff10:   nopl   0x6000808(%rax,%rax,1)       ;   {post_call_nop}

```

I checked that `A4` code follows the same pattern, not showing it here for brevity.

Note that none of this code checked the class init state at runtime. _So_, `A4` does not know to check if class was loaded, and just unconditionally traps. It would not help to remove traps in preload code, and it does not matter that interpreter loaded the class after the first deopt. A4 would trap regardless. This is why this proposed heuristics works: if we trapped from AP4, we are likely to trap from a similar code shape in A4.

I guess the better fix would be to make sure both AP4 and A4 versions are compiled with clinit-barriers everywhere?

shipilev:
Note the improvements in `User` time, these are the compilations we do not do. This somewhat reflects in end-to-end timings too.

`JavacBenchApp 1`:

```
Benchmark 1: build/linux-x86_64-server-release/images/jdk/bin/java -Xms64m -Xmx8g \
   -XX:+UseSerialGC -cp JavacBenchApp.jar -XX:AOTCache=app.aot JavacBenchApp 1

# Before
  Time (mean ± σ):     180.2 ms ±   3.5 ms    [User: 349.9 ms, System: 67.2 ms]
  Range (min … max):   173.5 ms … 184.9 ms    30 runs

# After
  Time (mean ± σ):     176.5 ms ±   2.3 ms    [User: 299.7 ms, System: 65.7 ms]
  Range (min … max):   173.4 ms … 181.0 ms    30 runs
```

`JavacBenchApp 50`:

```
Benchmark 1: build/linux-x86_64-server-release/images/jdk/bin/java -Xms64m -Xmx8g \
   -XX:+UseSerialGC -cp JavacBenchApp.jar -XX:AOTCache=app.aot JavacBenchApp 50

# Before
  Time (mean ± σ):     467.4 ms ±   8.3 ms    [User: 2150.0 ms, System: 172.9 ms]
  Range (min … max):   455.5 ms … 488.6 ms    30 runs

# After
  Time (mean ± σ):     453.7 ms ±   4.8 ms    [User: 1567.3 ms, System: 146.4 ms]
  Range (min … max):   446.7 ms … 468.7 ms    30 runs

```

shipilev:
OK, so here is my argument. 

Some uncommon traps are endemic to preload code. I _suspect_ the one in the disassembly above comes from `ciTypeFlow::StateVector::do_new`. Regardless, I believe that if preload code traps, then non-preload code would also trap in those conditions, since both versions were compiled in the same environment. 

When we start eliminating more traps in preload code, this heuristics would naturally go away for those cases. (I need to finish https://github.com/openjdk/leyden/pull/38, probably when G1 barriers work lands and we can re-evaluate the impact). But right now, this heuristics helps the compilation times _a lot_, see the comment above.

In new version, I have extended MDO to track preload decompiles exactly, so we do not mix with "normal" decompilations and profiles.

vnkozlov:
> Some uncommon traps are endemic to preload code. I suspect the one in the disassembly above could come from places like ciTypeFlow::StateVector::do_new. Regardless, I believe that if preload code traps, then non-preload code would also trap in those conditions, since both versions were compiled in the same environment.

Right. They are also compiled by the same compiler thread one after an other: [output.cpp#L3463](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/opto/output.cpp#L3463)
Note, `Reason_PrecompileForPreload` is used only for precompilation feature which we are not using for our workflow AFAICS.

shipilev:
I need to double-check the dynamics for some of the methods to make sure we are not delaying the T4 compilation unnecessarily when we bail when loading A4 code.

src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -1080,6 +1080,9 @@ void ciEnv::register_method(ciMethod* target,
       MethodData* mdo = method()->method_data();
       if (mdo != nullptr && _inc_decompile_count_on_failure) {
         mdo->inc_decompile_count();
+        if (nm != nullptr && nm->preloaded()) {
```
vnkozlov, line 1083: `nm` is not set yet here. You need to check `preload`

shipilev: Right. Actually, I think I need to invalidate the A4 SCC entry anyway, so I can just do that on trapping path, and avoid extending any MDOs. This also avoids any attempt to actually load A4 code, touching any queues, or interfering with compilation policies. I pushed a new version, going to measure its perf tomorrow.

src/hotspot/share/code/SCCache.cpp
```diff
@@ -4456,3 +4458,22 @@ SCCache::ReadingMark::~ReadingMark() {
     }
   }
 }
+
+void SCCache::block_loading(Method* method, CompLevel comp_level) {
```
vnkozlov, line 4462: The name is a little confusing. May be `invalidate_trapped_entry()`?

iwanowww:
> Right. They are also compiled by the same compiler thread one after an other: [output.cpp#L3463](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/opto/output.cpp#L3463)
> Note, Reason_PrecompileForPreload is used only for precompilation feature which we are not using for our workflow AFAICS.

Actually, it's not how it works in 1-step workflow. `Reason_Precompile` and `Reason_PrecompileForPreload` are used for populating AOT code archive during assembly phase (see [precompiler.cpp#L138](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/compiler/precompiler.cpp#L138)).

vnkozlov:
> [precompiler.cpp#L138](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/compiler/precompiler.cpp#L138)

vnkozlov:
> > Right. They are also compiled by the same compiler thread one after an other: [output.cpp#L3463](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/opto/output.cpp#L3463)
> > Note, Reason_PrecompileForPreload is used only for precompilation feature which we are not using for our workflow AFAICS.
> 
> Actually, it's not how it works in 1-step workflow. `Reason_Precompile` and `Reason_PrecompileForPreload` are used for populating AOT code archive during assembly phase (see [precompiler.cpp#L138](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/compiler/precompiler.cpp#L138)).

Okay, I see it now: [metaspaceShared.cpp#L1116](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/cds/metaspaceShared.cpp#L1116)

Which means we compile "preload" and "normal" AOT code by different tasks and they may have different MDO due to pollution.

iwanowww:
I still feel a bit uneasy about the proposed change. 

> Regardless, I believe that if preload code traps, then non-preload code would also trap in those conditions, since both versions were compiled in the same environment.

Unless it caused by a divergence in behavior between training and production runs, I'd attribute it to a bug in our code where we misinterpret training data (e.g., excluding the class from the archive, not preloading class in all loaders, not preresolving relevant CP entry, etc). 

Before making a decision, I'd like to understand what's the real cause of the trap at `new NameImpl(this, string)` in your case.

```
[0.047s][debug  ][deoptimization] cid= 333     level=4 com.sun.tools.javac.util.StringNameTable::fromString(Ljava/lang/String;)Lcom/sun/tools/javac/util/Name; trap_bci=28 unloaded reinterpret pc=0x000071945803a2c0 relative_pc=0x00000000000003c0
```

iwanowww:
> Which means we compile "preload" and "normal" AOT code by different tasks and they may have different MDO due to pollution.

There should be no divergence in profiling data between compilations unless there's Java code execution happening concurrently. But by the time `Precompiler::compile_cached_code()` is invoked, there's no Java code executed anymore.

vnkozlov:
We generate more AOT code with this. With one task request before I don't generate second "normal" version if code does not have init barriers. Now we generate 2 copies of the same code.  May be this is fine performance vise but not space saving.

iwanowww:
For `Reason_Precompile` and `Reason_PrecompileForPreload` C2 compilations, MDOs are unconditionally taken from training data (see [ciMethod::method_data()](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/ci/ciMethod.cpp#L1062)).

vnkozlov:
We may also generate 2 version of tier2 because we generate them for limited_profiling and full_profiling.

vnkozlov:
> For `Reason_Precompile` and `Reason_PrecompileForPreload` C2 compilations, MDOs are unconditionally taken from training data (see [ciMethod::method_data()](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/ci/ciMethod.cpp#L1062)).

That is good. Then the assumption that both code versions have same traps is correct.

iwanowww:
> We generate more AOT code with this. With one task request before I don't generate second "normal" version if code does not have init barriers. Now we generate 2 copies of the same code. May be this is fine performance vise but not space saving.

It sounds like a good optimization, but now there are more divergences in generated code (reduction in uncommon trap usage). In a future, I'd expect more distinctions to be introduced.

iwanowww:
> We may also generate 2 version of tier2 because we generate them for limited_profiling and full_profiling.

Good point. That's a bug.

shipilev:
Closing in favor of [JDK-8349814](https://bugs.openjdk.org/browse/JDK-8349814), which made uncommon traps in preload code much less visible. We can re-open this if we find this is still useful.

PR #49, opened 2025-03-14
8352042: [leyden] Parallel precompilation

shipilev:
For simplicity reasons, our current precompilation code uses effectively single-threaded blocking compiles. Even for simple cases, this makes assembly phase remarkably long. This does not scale well. We need to do precompilations in parallel. 

Unfortunately, I haven't found a good way to check that all async compilations are completed, so I introduced a new one.

Additional testing:
 - [x] Ad-hoc precompilation tests

### DIFF
--- a/src/hotspot/share/compiler/compileBroker.cpp
+++ b/src/hotspot/share/compiler/compileBroker.cpp
@@ -2045,6 +2045,10 @@ void CompileBroker::wait_for_completion(CompileTask* task) {
   }
 }
 
+void CompileBroker::wait_for_no_active_tasks() {
+  CompileTask::wait_for_no_active_tasks();
+}
+
 /**
  * Initialize compiler thread(s) + compiler object(s). The postcondition
  * of this function is that the compiler runtimes are initialized and that
--- a/src/hotspot/share/compiler/compileBroker.cpp
+++ b/src/hotspot/share/compiler/compileBroker.cpp
@@ -1755,9 +1755,7 @@ nmethod* CompileBroker::compile_method(const methodHandle& method, int osr_bci,
     }
     bool is_blocking = ReplayCompiles                                             ||
                        !directive->BackgroundCompilationOption                    ||
-                       (PreloadBlocking && (compile_reason == CompileTask::Reason_Preload)) ||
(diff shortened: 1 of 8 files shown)


shipilev:
```
$ ./run-leyden.sh 

# Before

Generated source code for 10001 classes and compiled them in 19221 ms
AOTConfiguration recorded: app.aotconf

real	0m20.195s
user	0m51.408s
sys	0m1.466s

Reading AOTConfiguration app.aotconf and writing AOTCache app.aot
AOTCache creation is complete: app.aot

real	0m25.396s  ; <--- no parallelism
user	0m25.212s
sys	0m0.696s


# After

Generated source code for 10001 classes and compiled them in 19320 ms
AOTConfiguration recorded: app.aotconf

real	0m20.228s
user	0m50.004s
sys	0m1.419s

Reading AOTConfiguration app.aotconf and writing AOTCache app.aot
AOTCache creation is complete: app.aot

real	0m3.672s   ; <--- 6.6x parallelism
user	0m24.634s
sys	0m0.715s
```

rose00:
Really nice, thanks.  What motivated the type change on `CompileTaskAlloc_lock`?  Some kind of end-of-time effect that broke the mutex grab?

shipilev:
> What motivated the type change on `CompileTaskAlloc_lock`? Some kind of end-of-time effect that broke the mutex grab?

We now need notifications, so the lock is no longer a pure `Mutex`, it is `Monitor` now.

iwanowww (approved):
Thanks for taking care of it, Aleksey. 

Overall, looks good. `CompileTask::_active_tasks` feels a bit ad-hoc, but IMO it's fine for leyden-premain.

Also, since AOT code preparation is a synchronous task during assembly phase, it makes sense to revise how many compiler threads are used.

shipilev:
> Also, since AOT code preparation is a synchronous task during assembly phase, it makes sense to revise how many compiler threads are used.

I had the same thought, but the effects I measured were not compelling. So I left the number of compiler threads to the default dynamic policy, which seems to work well.

shipilev:
Thanks for reviews, I am going to integrate now. If there are follow-ups, we can handle them separately.

/integrate

PR #52, opened 2025-04-02
8356572: [premain] Add management MXBean for AOT

macarte:
Adding AOTCacheMXBean as per JBS issue

```
    /**
      * Returns the string representing the current AOT mode of
      * operation.
      *
      * @return the string representing the current AOT mode.
      */
     public String getMode();

     /**
      * Tests if a recording is in progress.
      *
      * @return {@code true} if a recording is in progress; {@code false} otherwise.
      */
     public boolean isRecording();

     /**
      * If a recording is in progress or has been completed, then returns the duration in milliseconds
      *
      * @return duration of the recording in milliseconds.
      */
     public long getRecordingDuration();

     /**
      * If a recording is in progress, then stops the recording.
      *
      * @return {@code true} if a recording was stopped; {@code false} otherwise.
      */
     public boolean endRecording();

```

---------
### Progress

### Issue
 * [JDK-8356572](https://bugs.openjdk.org/browse/JDK-8356572): [premain] Add management MXBean for AOT (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/52/head:pull/52` \
`$ git checkout pull/52`

`$ git checkout pull/52` \
`$ git pull https://git.openjdk.org/leyden.git pull/52/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 52`

`$ git pr show -t 52`

</details>

<a href="https://git.openjdk.org/leyden/pull/52.diff">https://git.openjdk.org/leyden/pull/52.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/52#issuecomment-2813451088)
</details>


iklam:
For a test case, I think you can make a copy of the following test case, but modify the MyTestApp to call the MX Bean instead. 

https://github.com/openjdk/leyden/blob/premain/test/hotspot/jtreg/runtime/cds/appcds/leyden/EndTrainingOnMethodEntry.java

macarte:
@iklam - added test, as well as testing it stops training, it also tests that all bean methods work as expected in the various aot modes

AlanBateman:
AOT is a JDK feature rather than standard feature so you might want to think of making it JDK specific, in jdk.management, rather than as a standard MXBean.

iklam:
> AOT is a JDK feature rather than standard feature so you might want to think of making it JDK specific, in jdk.management, rather than as a standard MXBean.

Is there any difference in terms of visibility between java.management vs jdk.management? Aer both modules visible by default in a standard JDK build?

test/hotspot/jtreg/runtime/cds/appcds/leyden/EndTrainingWithAOTMXBean.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        @Override
+        public String classpath(RunMode runMode) {
+            return appJar;
+        }
+
+        public String[] vmArgs(RunMode runMode) {
+            return new String[] {
+                "-Xlog:cds+class=debug"
```
iklam, line 77: For completeness, I think we should add `"--add-modules=java.management"` to here.

macarte:
@AlanBateman / @iklam - thank you for the review and recommendations.  I've updated this PR with changes to move the MXBean from java.management to jdk.management

iklam:
The module names look good. Perhaps we can have a more descriptive name than AOTMXBean? @AlanBateman any suggestions?

AlanBateman:
> The module names look good. Perhaps we can have a more descriptive name than AOTMXBean? @AlanBateman any suggestions?

Something like "AheadOfTimeCacheMXBean", registered with the MBeanServer as "jdk.management:type=AheadOfTimeCache", might be closer to what you want.

An enum could be used to model the modes and that would give you a place to specify them too. Right now the prototype uses a string "representing the current AOT mode" so you would need to search around to know what the modes are. 

The operation to stop the recording is currently named "endRecording" so there is a mix of "stop" and "end", pick one.

The recording duration attribute (getRecordingDuration method) will need to say more. When does recording start, so it this VM uptime?

vnkozlov:
@AlanBateman  We don't use AheadOfTime in our flags or class names. We use well known acronym AOT. I think we should be consistent everywhere.

AlanBateman:
Okay, but I think the issue with "AOTMXBean" is that it's a management interface AOT cache and recording status, there may be other management interfaces for other aspects to AOT in the future.

mbreinhold:
`AheadOfTimeCacheMXBean` is quite a mouthful. “AOT” is a reasonably standard term of art, and will be familar to anyone using the HotSpot AOT CLI options. So why not `AOTCacheMXBean`?

macarte:
I've renamed AOT* to AOTCache*; I'm sure we'll go through another round of reviews before it exits premain

vnkozlov:
Please also update JBS and PR's Subject to new name.

macarte:
Linked to new issue: https://bugs.openjdk.org/browse/JDK-8356572

macarte:
> Please also update JBS and PR's Subject to new name.

I hadn't created an issue, but I've added one now and updated the PR title to reflect that

vnkozlov (approved):
Good.

PR #53, opened 2025-04-10
8352042: [leyden] Improve parallel recompilation parallelism

shipilev:
Parallel precompilation ([JDK-8352042](https://bugs.openjdk.org/browse/JDK-8352042)) uses the default compiler thread policy, that targets the normally executing applications. In assembly phase, we can actually use all threads available on the machine to complete it as quickly as possible.


vnkozlov (approved):
Seems fine.

shipilev:
Thanks!

/integrate

PR #60, opened 2025-04-23
8355352: [premain] rename AOT Code classes and logging tags

vnkozlov:
Rename AOT Code classes and logging tags to match changes for mainline: [#24740](https://github.com/openjdk/jdk/pull/24740)

This is first part of changes in premain branch.  AOT code flags renaming and more code changes will be done in separate PRs.

Tested with premain-tier1

### DIFF
--- a/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
@@ -28,9 +28,6 @@
 #include "ci/ciEnv.hpp"
 #include "ci/ciUtilities.hpp"
 #include "code/compiledIC.hpp"
-#if INCLUDE_CDS
-#include "code/SCCache.hpp"
-#endif
 #include "compiler/compileTask.hpp"
 #include "compiler/disassembler.hpp"
 #include "compiler/oopMap.hpp"
@@ -699,7 +696,7 @@ void MacroAssembler::set_last_Java_frame(Register last_java_sp,
 }
 
 static inline bool target_needs_far_branch(address addr) {
-  if (SCCache::is_on_for_write()) {
+  if (AOTCodeCache::is_on_for_write()) {
     return true;
   }
   // codecache size <= 128M
@@ -886,7 +883,7 @@ void MacroAssembler::call_VM_helper(Register oop_result, address entry_point, in
 
 // Check the entry target is always reachable from any branch.
 static bool is_always_within_branch_range(Address entry) {
-  if (SCCache::is_on_for_write()) {
+  if (AOTCodeCache::is_on_for_write()) {
     return false;
   }
   const address target = entry.target();
@@ -3269,7 +3266,7 @@ void MacroAssembler::stop(const char* msg) {
   // ExternalAddress enables saving and restoring via the code cache
   lea(c_rarg0, ExternalAddress((address) msg));
   dcps1(0xdeae);
-  SCCache::add_C_string(msg);
+  AOTCodeCache::add_C_string(msg);
 }
 
 void MacroAssembler::unimplemented(const char* what) {
@@ -3367,7 +3364,7 @@ void MacroAssembler::subw(Register Rd, Register Rn, RegisterOrConstant decrement
 void MacroAssembler::reinit_heapbase()
 {
   if (UseCompressedOops) {
-    if (Universe::is_fully_initialized() && !SCCache::is_on_for_write()) {
+    if (Universe::is_fully_initialized() && !AOTCodeCache::is_on_for_write()) {
       mov(rheapbase, CompressedOops::base());
     } else {
       lea(rheapbase, ExternalAddress(CompressedOops::base_addr()));
@@ -5737,8 +5734,8 @@ void MacroAssembler::load_byte_map_base(Register reg) {
   // Strictly speaking the byte_map_base isn't an address at all, and it might
   // even be negative. It is thus materialised as a constant.
 #if INCLUDE_CDS
-  if (SCCache::is_on_for_write()) {
-    // SCA needs relocation info for card table base
+  if (AOTCodeCache::is_on_for_write()) {
+    // AOT code needs relocation info for card table base
     lea(reg, ExternalAddress(reinterpret_cast<address>(byte_map_base)));
   } else {
 #endif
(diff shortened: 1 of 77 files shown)


src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -28,9 +28,6 @@
 #include "ci/ciEnv.hpp"
 #include "ci/ciUtilities.hpp"
 #include "code/compiledIC.hpp"
-#if INCLUDE_CDS
```
shipilev, line 31: Did you mean to only remove `INCLUDE_CDS` , and add the `#include` for `autoCodeCache.hpp`?

src/hotspot/cpu/x86/gc/g1/g1BarrierSetAssembler_x86.cpp
```diff
@@ -24,7 +24,7 @@
 
 #include "asm/macroAssembler.inline.hpp"
 #if INCLUDE_CDS
-#include "code/SCCache.hpp"
```
shipilev, line 27: While you are at it, remove `INCLUDE_CDS` guards around includes like these?

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -4592,18 +4592,18 @@ address SCAddressTable::address_for_id(int idx) {
...
-  fatal("Incorrect id %d for SCA table", id);
+  fatal("Incorrect id %d for AOT Code Cache addresses table", id);
   return nullptr;
 }
 
-int SCAddressTable::id_for_address(address addr, RelocIterator reloc, CodeBuffer* buffer) {
+int AOTCodeAddressTable::id_for_address(address addr, RelocIterator reloc, CodeBuffer* buffer) {
+  if (!_extrs_complete) {
+    fatal("AOT Code Cache VM runtime addresses table is not complete");
+  }
```
shipilev, line 4602: We sure this move makes sense?

src/hotspot/share/compiler/compilationPolicy.cpp
```diff
@@ -1091,7 +1091,7 @@ bool CompilationPolicy::compare_methods(Method* x, Method* y) {
 }
 
 bool CompilationPolicy::compare_tasks(CompileTask* x, CompileTask* y) {
-  assert(!x->is_scc() && !y->is_scc(), "SC tasks are not expected here");
+  assert(!x->is_aot() && !y->is_aot(), "SC tasks are not expected here");
```
shipilev, line 1094: "AOT tasks are not expected here"?

src/hotspot/share/compiler/compileBroker.cpp
```diff
@@ -3198,7 +3198,7 @@ void CompileBroker::print_times(bool per_compiler, bool aggregate) {
 
   if (StoreCachedCode || LoadCachedCode) { // Check flags because SC cache could be closed already
     tty->cr();
-    SCCache::print_timers_on(tty);
+    AOTCodeCache::print_timers_on(tty);
```
shipilev, line 3201: There is a mention of "SC cache" a few lines above.

src/hotspot/share/compiler/compileTask.cpp
```diff
@@ -22,7 +22,6 @@
  *
  */
 
-#include "code/SCCache.hpp"
```
shipilev, line 25: This compilation unit uses `AOTCode*` stuff, so the include for `aotCodeCache.hpp` should be here?

shipilev (commented):
Looks reasonable. Nits below. Please enable GHA for this PR and make sure it is green.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
vnkozlov: `macroAssembler_aarch64.hpp` already has #include.

src/hotspot/cpu/x86/gc/g1/g1BarrierSetAssembler_x86.cpp
vnkozlov: Will do.

src/hotspot/share/code/aotCodeCache.cpp
vnkozlov: `id_for_address()` is called for processing relocation info in compiled code (stubs, adapters, code). We should not do that until we initialized the table for VM's runtime addresses at least! We do that initialization very early in `SCCache::init2()` which is called just after `universe_init()`.

src/hotspot/share/compiler/compilationPolicy.cpp
vnkozlov: Will fix.

src/hotspot/share/compiler/compileBroker.cpp
vnkozlov: Will fix

src/hotspot/share/code/aotCodeCache.cpp
shipilev: OK!

src/hotspot/share/compiler/compileTask.cpp
vnkozlov: The only reference is `AOTCodeEntry*`  which is declared in `compileTask.hpp`. No methods from it are called.

shipilev: Right, OK.

vnkozlov:
Thank you, @shipilev, for looking on changes. I will address issues you pointed.

But I screwed up by using leyden/premain branch directly with these changes and not in my fork. That is why there is no GHA testing (and I can't activate it).

I will submit new PR from my fork.

src/hotspot/share/code/aotCodeCache.cpp
vnkozlov: I looked on my original changes and I think I placed the check there because `int id = -1;` was first instruction.
But I will run more testing to see when `-1` is passed to make sure this is fine.

vnkozlov:
Closing this PR in favor of new correct PR (from my fork and GHA testing): https://github.com/openjdk/leyden/pull/62

PR #63, opened 2025-04-28
Cleaner GHA build of hermetic-java-runtime branch

jerboaa:
Please review this simple PR which fixes build issues seen on GHA. Before this PR we see build failures in various Linux cross-builds and windows builds. This PR should fix those. Note: There are still GHA test failures, but at least we can build the branch on the main 3 platforms.

Thoughts?

### DIFF
--- a/src/hotspot/os_cpu/linux_aarch64/threadLS_linux_aarch64.S
+++ b/src/hotspot/os_cpu/linux_aarch64/threadLS_linux_aarch64.S
@@ -31,10 +31,10 @@
 DECLARE_FUNC(_ZN10JavaThread25aarch64_get_thread_helperEv):
         hint #0x19      // paciasp
         stp x29, x30, [sp, -16]!
-        adrp x0, :tlsdesc:_ZN17HotspotBaseThread12_thr_currentE
-        ldr x1, [x0, #:tlsdesc_lo12:_ZN17HotspotBaseThread12_thr_currentE]
-        add x0, x0, :tlsdesc_lo12:_ZN17HotspotBaseThread12_thr_currentE
-        .tlsdesccall _ZN17HotspotBaseThread12_thr_currentE
+        adrp x0, :tlsdesc:_ZN6Thread12_thr_currentE
+        ldr x1, [x0, #:tlsdesc_lo12:_ZN6Thread12_thr_currentE]
+        add x0, x0, :tlsdesc_lo12:_ZN6Thread12_thr_currentE
+        .tlsdesccall _ZN6Thread12_thr_currentE
         blr x1
         mrs x1, tpidr_el0
         add x0, x1, x0
--- a/src/hotspot/share/utilities/globalDefinitions.hpp
+++ b/src/hotspot/share/utilities/globalDefinitions.hpp
@@ -1026,10 +1026,6 @@ enum JavaThreadState {
   _thread_max_state         = 12  // maximum thread state+1 - used for statistics allocation
 };
 
-// Redefine Thread as BaseThread to avoid duplicate symbol issues for
-// JDK static builds. See runtime/thread.hpp for Thread class.
-#define Thread HotspotBaseThread
-
 enum LockingMode {
   // Use only heavy monitors for locking
   LM_MONITOR     = 0,
--- a/src/java.base/share/native/launcher/main.c
+++ b/src/java.base/share/native/launcher/main.c
@@ -36,6 +36,13 @@
 
 #include <limits.h>
 
+// Windows doesn't have PATH_MAX. It's MAX_PATH instead.
+#ifdef _WIN32
+#ifndef PATH_MAX
+#define PATH_MAX MAX_PATH
+#endif
+#endif
+
 static jboolean read_u8(FILE *f, jboolean is_little_endian, unsigned long long* res) {
     unsigned char* v = (unsigned char*)res;
     if (is_little_endian) {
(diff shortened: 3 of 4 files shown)


src/hotspot/share/utilities/globalDefinitions.hpp
```diff
@@ -1026,10 +1026,6 @@ enum JavaThreadState {
   _thread_max_state         = 12  // maximum thread state+1 - used for statistics allocation
 };
 
-// Redefine Thread as BaseThread to avoid duplicate symbol issues for
-// JDK static builds. See runtime/thread.hpp for Thread class.
-#define Thread HotspotBaseThread
```
jianglizhou, line 1031: This change and the one in src/hotspot/os_cpu/linux_aarch64/threadLS_linux_aarch64.S are related to https://bugs.openjdk.org/browse/JDK-8311846. There were discussions on https://github.com/openjdk/jdk/pull/17456 (closed without integrating), but we don't yet have a acceptable solution moving forward.

Should we leave the patch in the branch until we identify a proper solution?

src/java.base/share/native/launcher/main.c
```diff
@@ -36,6 +36,13 @@
 
 #include <limits.h>
 
+// Windows doesn't have PATH_MAX. It's MAX_PATH instead.
+#ifdef _WIN32
```
jianglizhou, line 40: Thanks. Checking https://learn.microsoft.com/en-us/cpp/c-language/cpp-integer-limits?view=msvc-170, PATH_MAX is not defined for Windows limits.h.

src/hotspot/share/utilities/globalDefinitions.hpp
jerboaa: Is this duplicate symbol issue still a problem? If so, how would I be able to reproduce? This breaks cross compiled code. For example see: https://github.com/jerboaa/leyden/actions/runs/14613421232/job/40996157052

jianglizhou: > Is this duplicate symbol issue still a problem? If so, how would I be able to reproduce? This breaks cross compiled code. For example see: https://github.com/jerboaa/leyden/actions/runs/14613421232/job/40996157052

The `Thread` duplicate symbol issue exists when statically linking with any application native code and native dependencies with a symbol also defined as `Thread`. We ran into the issue when prototyping on JDK 11. 

```
In file included from /home/runner/work/leyden/leyden/src/hotspot/share/cds/aotClassFilter.hpp:29,
                 from /home/runner/work/leyden/leyden/src/hotspot/share/cds/aotClassFilter.cpp:25:
/home/runner/work/leyden/leyden/src/hotspot/share/cds/aotClassFilter.cpp: In destructor ‘AOTClassFilter::FilterMark::~FilterMark()’:
/home/runner/work/leyden/leyden/src/hotspot/share/cds/aotClassFilter.cpp:42:53: error: comparison between distinct pointer types ‘Thread*’ and ‘HotspotBaseThread*’ lacks a cast
   42 |   assert(_current_mark == this && _filtering_thread == Thread::current(), "sanity");
```

Ok, we also ran into aotClassFilter.* related build failure when syncing with JDK mainline internally in early April. I've resolved the build issue in our code base. I'll push an update to hermetic-java-runtime branch to address this build issue shortly. Thanks for noticing it.

jianglizhou: I pushed https://git.openjdk.org/leyden/commit/c93d26d944b65b955a78c4227776bb0fc1d72363.

jerboaa: OK. I've merged in the latest updates and reverted the change in `src/hotspot/share/utilities/globalDefinitions.hpp`. Let's see how GHA looks.

src/hotspot/os_cpu/linux_aarch64/threadLS_linux_aarch64.S
jianglizhou, line 1: The src/hotspot/os_cpu/linux_aarch64/threadLS_linux_aarch64.S also needs to be reverted. It causes the linux-cross-build aarch64 build failures:

```
...
/usr/lib/gcc-cross/aarch64-linux-gnu/10/../../../../aarch64-linux-gnu/bin/ld: /home/runner/work/leyden/leyden/src/hotspot/os_cpu/linux_aarch64/threadLS_linux_aarch64.S:38: undefined reference to `Thread::_thr_current'
collect2: error: ld returned 1 exit status
* For target hotspot_variant-server_libjvm_objs_BUILD_LIBJVM_run_ld:
/usr/lib/gcc-cross/aarch64-linux-gnu/10/../../../../aarch64-linux-gnu/bin/ld: /home/runner/work/leyden/leyden/build/linux-aarch64/hotspot/variant-server/libjvm/objs/threadLS_linux_aarch64.o: in function `JavaThread::aarch64_get_thread_helper()':
/home/runner/work/leyden/leyden/src/hotspot/os_cpu/linux_aarch64/threadLS_linux_aarch64.S:34: undefined reference to `Thread::_thr_current'
...
```

jerboaa: Done.

jianglizhou (approved):

src/hotspot/share/utilities/globalDefinitions.hpp
magicus: @jerboaa The long-term plan is to handle this issue by using the linker to turn non-exported public symbols into private symbols before generating static libraries. This works on linux and macos, but I have not yet gotten it to work properly on Windows, and there is only a proof-of-concept sketch on how to get it to work on AIX.

jerboaa: OK.

PR #64, opened 2025-04-29
8355915: [leyden] Crash in MDO clearing the unloaded array type

shipilev:
Caught this when doing benchmarks with Spring Boot. See the bug for reproducer.

The lifecycle of array types of `T` is bound to the lifecycle of `T` themselves. So the fix does similar thing to `is_excluded` and other code: when we encounter the `T[]`, we ask if its bottom component classes is not yet initialized. This fits nicely with our current uses, which skip such classes, and ever touch their (potentially nullptr) CLDs.

Additional testing:
 - [x] Ad-hoc testing with most recent spring-boot-petclinic

---------
### Progress

### Issue
 * [JDK-8355915](https://bugs.openjdk.org/browse/JDK-8355915): [leyden] Crash in MDO clearing the unloaded array type (**Bug** - P4)

### Reviewers
 * [Vladimir Ivanov](https://openjdk.org/census#vlivanov) (@iwanowww - Committer) ⚠️ Review applies to [ab91db2d](https://git.openjdk.org/leyden/pull/64/files/ab91db2dd72291dbb10bb0eb36f3b9e7c4b06839)
 * [Igor Veresov](https://openjdk.org/census#iveresov) (@veresov - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/64/head:pull/64` \
`$ git checkout pull/64`

`$ git checkout pull/64` \
`$ git pull https://git.openjdk.org/leyden.git pull/64/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 64`

`$ git pr show -t 64`

</details>

<a href="https://git.openjdk.org/leyden/pull/64.diff">https://git.openjdk.org/leyden/pull/64.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/64#issuecomment-2839262311)
</details>

### DIFF
--- a/src/hotspot/share/ci/ciMethodData.cpp
+++ b/src/hotspot/share/ci/ciMethodData.cpp
@@ -64,10 +64,6 @@ static bool is_klass_loaded(Klass* k) {
   return true;
 }
 
-static bool has_cld(Klass *k) {
-  return k != nullptr && k->class_loader_data() != nullptr;
-}
-
 // Check for entries that reference an unloaded method
 class PrepareExtraDataClosure : public CleanExtraDataClosure {
   MethodData*            _mdo;
@@ -83,7 +79,7 @@ public:
 
   bool is_live(Method* m) {
     Klass* holder = m->method_holder();
-    if (!has_cld(holder) || !holder->is_loader_alive() || !is_klass_loaded(holder)) {
+    if (holder == nullptr || !holder->is_loader_present_and_alive() || !is_klass_loaded(holder)) {
       return false;
     }
     if (CURRENT_ENV->cached_metadata(m) == nullptr) {
@@ -318,7 +314,7 @@ bool ciMethodData::load_data() {
 void ciReceiverTypeData::translate_receiver_data_from(const ProfileData* data) {
   for (uint row = 0; row < row_limit(); row++) {
     Klass* k = data->as_ReceiverTypeData()->receiver(row);
-    if (has_cld(k) && is_klass_loaded(k)) {
+    if (k != nullptr && k->is_loader_present_and_alive() && is_klass_loaded(k)) {
       if (k->is_loader_alive()) {
         ciKlass* klass = CURRENT_ENV->get_klass(k);
         set_receiver(row, klass);
@@ -336,7 +332,7 @@ void ciTypeStackSlotEntries::translate_type_data_from(const TypeStackSlotEntries
   for (int i = 0; i < number_of_entries(); i++) {
     intptr_t k = entries->type(i);
     Klass* klass = (Klass*)klass_part(k);
-    if (!has_cld(klass) || !klass->is_loader_alive() || !is_klass_loaded(klass)) {
+    if (klass == nullptr || !klass->is_loader_present_and_alive() || !is_klass_loaded(klass)) {
       // With concurrent class unloading, the MDO could have stale metadata; override it
       TypeStackSlotEntries::set_type(i, TypeStackSlotEntries::with_status((Klass*)nullptr, k));
     } else {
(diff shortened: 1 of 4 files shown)


iwanowww (approved):
Looks good.

shipilev:
Ha! There are tier1 failures even without this fix. Going to try and fix them at the same time.

shipilev:
Testing shows the fix is incomplete. 

Looks like when `T[]` is unloaded, we cannot trust its `bottom_klass` is not garbage. I also notice that in several other places, e.g. in `ciReceiverTypeData::translate_receiver_data_from`, we do check if CLD is null. So the more robust fix is to do CLD checks in MDO clearing as well. `is_excluded` is now also protected by these CLD checks.

See new commit. This one seems to pass the tests locally.

shipilev:
I now see @veresov had a similar trouble: https://github.com/shipilev/leyden/commit/a77cc17caa039a862b62cd64c67e35e666b66a48. I think we can both a bit cleaner by introducing an utility method straight in `Klass`.

shipilev:
Are we good with this? @iwanowww, @veresov?

veresov (approved):
Looks ok

shipilev:
Thanks! Igor, I think you want to mix this into your persistent profiles PR.

/integrate

veresov:
> Thanks! Igor, I think you want to mix this into your persistent profiles PR.
> 
> /integrate

Yup, I'll port it.

src/hotspot/share/ci/ciMethodData.cpp
```diff
@@ -83,7 +79,7 @@ class PrepareExtraDataClosure : public CleanExtraDataClosure {
 
   bool is_live(Method* m) {
     Klass* holder = m->method_holder();
-    if (!has_cld(holder) || !holder->is_loader_alive() || !is_klass_loaded(holder)) {
+    if (holder == nullptr || !holder->is_loader_present_and_alive() || !is_klass_loaded(holder)) {
```
iwanowww, line 82: Why do we need these checks on CI side?

veresov: Wasn't it you who put it there? :)

iwanowww: Nevermind, it turns out to be just a refactoring:`holder == nullptr || !holder->is_loader_present_and_alive()` is equivalent to the original check `!has_cld(holder) || !holder->is_loader_alive()`.

src/hotspot/share/oops/methodData.cpp
```diff
@@ -349,7 +349,7 @@ void TypeStackSlotEntries::clean_weak_klass_links(bool always_clean) {
       if (!always_clean && k->is_instance_klass() && InstanceKlass::cast(k)->is_not_initialized()) {
         continue; // skip not-yet-initialized classes // TODO: maybe clear the slot instead?
       }
-      if (always_clean || !k->is_loader_alive() || is_excluded(k)) {
+      if (always_clean || !k->is_loader_present_and_alive() || is_excluded(k)) {
```
iwanowww, line 352: In the longer term, the thing which concerns me about it is that both `is_excluded()` (and `is_klass_loaded()` in `ciMethodData.cpp`) are deliberately limited to Leyden-related modes while `k->is_loader_present_and_alive()` is performed unconditionally across all execution modes. Seeing a Klass with a missing CLD during normal execution mode is a sign of a bug in some other place.

src/hotspot/share/ci/ciMethodData.cpp
```diff
@@ -318,7 +314,7 @@ bool ciMethodData::load_data() {
 void ciReceiverTypeData::translate_receiver_data_from(const ProfileData* data) {
   for (uint row = 0; row < row_limit(); row++) {
     Klass* k = data->as_ReceiverTypeData()->receiver(row);
-    if (has_cld(k) && is_klass_loaded(k)) {
+    if (k != nullptr && k->is_loader_present_and_alive() && is_klass_loaded(k)) {
```
veresov, line 317: This little part is actually incorrect. It should be just `k->class_loader_data() != nullptr` because there are different branch before for handling if `is_loader_alive()` and if not. I'll push the fix.

PR #67, opened 2025-05-16
8347830: [premain] UseCompatibleCompressedOops is broken after merging with mainline

vnkozlov:
The issue was that we did not allocate space "noaccess prefix" before heap for read-only page when we setup type of encoding  with `UseCompatibleCompressedOops` flag. I update code which reserves heap for compressed oops to take into account this flag.

An other issue was that `UseCompatibleCompressedOops` flag was set during CDS arguments consistency checkup. But `UseCompressedOops` flag is set by GC ergonomics. I moved code to CDS ergonomics setting method.

During testing I hit assert that AOT code address table is missing some code blob address. But it was actually card table base address which pointed inside CodeCache. I added special case to check for it early.

Tested premain-tier1.

### DIFF
--- a/src/hotspot/share/cds/cdsConfig.cpp
+++ b/src/hotspot/share/cds/cdsConfig.cpp
@@ -120,6 +120,26 @@ void CDSConfig::ergo_initialize() {
   if (!is_dumping_heap()) {
     _is_dumping_full_module_graph = false;
   }
+
+#ifdef _LP64
+  //
+  // By default, when using AOTClassLinking, use the CompressedOops::HeapBasedNarrowOop
+  // mode so that AOT code can be always work regardless of runtime heap range.
+  //
+  // If you are *absolutely sure* that the CompressedOops::mode() will be the same
+  // between training and production runs (e.g., if you specify -Xmx128m for
+  // both training and production runs, and you know the OS will always reserve
+  // the heap under 4GB), you can explicitly disable this with:
+  //     java -XX:+UnlockDiagnosticVMOptions -XX:-UseCompatibleCompressedOops ...
+  // However, this is risky and there's a chance that the production run will be slower
+  // than expected because it is unable to load the AOT code cache.
+  //
+  if (UseCompressedOops && AOTCodeCache::is_caching_enabled()) {
+    FLAG_SET_ERGO_IF_DEFAULT(UseCompatibleCompressedOops, true);
+  } else if (!FLAG_IS_DEFAULT(UseCompatibleCompressedOops)) {
+    FLAG_SET_ERGO(UseCompatibleCompressedOops, false);
+  }
+#endif // _LP64
 }
 
 const char* CDSConfig::default_archive_path() {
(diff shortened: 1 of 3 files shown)


vnkozlov:
@ashu-mehra and @adinn please look.

src/hotspot/share/cds/cdsConfig.cpp
```diff
@@ -120,6 +120,26 @@ void CDSConfig::ergo_initialize() {
...
+  // both training and production runs, and you know the OS will always reserve
+  // the heap under 4GB), you can explicitly disable this with:
+  //     java -XX:+UnlockDiagnosticVMOptions -XX:-UseCompatibleCompressedOops ...
+  // However, this is risky and there's a chance that the production run will be slower
+  // than expected because it is unable to load the AOT code cache.
+  //
+  if (UseCompressedOops && AOTCodeCache::is_caching_enabled()) {
+    FLAG_SET_ERGO_IF_DEFAULT(UseCompatibleCompressedOops, true);
+  } else if (!FLAG_IS_DEFAULT(UseCompatibleCompressedOops)) {
+    FLAG_SET_ERGO(UseCompatibleCompressedOops, false);
```
ashu-mehra, line 140: Why do we need to set it to false explicitly here? If it is not default shouldn't we use whatever value the user specified?

src/hotspot/share/memory/memoryReserver.cpp
```diff
@@ -549,8 +549,17 @@ ReservedHeapSpace HeapReserver::Instance::reserve_compressed_oops_heap(const siz
   const size_t attach_point_alignment = lcm(alignment, os_attach_point_alignment);
 
   char* aligned_heap_base_min_address = align_up((char*)HeapBaseMinAddress, alignment);
-  size_t noaccess_prefix = ((aligned_heap_base_min_address + size) > (char*)OopEncodingHeapMax) ?
-    noaccess_prefix_size : 0;
+  char* heap_end_address = aligned_heap_base_min_address + size;
+
+  bool unscaled  = false;
+  bool zerobased = false;
+  bool heapbased = UseCompatibleCompressedOops;
```
ashu-mehra, line 556: Use of `heapbased` is a bit confusing to me. Can we restructure it like this:
```
bool unscaled = false;
bool zerobased = false;
if (UseCompatibleCompressedOops) {
  unscaled = (heap_end_address <= (char*)UnscaledOopHeapMax);
  zerobased = (heap_end_address <= (char*)OopEncodingHeapMax);
}
size_t noaccess_prefix = (UseCompatibleCompressedOops || !zerobased) ? noaccess_prefix_size : 0;
```

src/hotspot/share/cds/cdsConfig.cpp
vnkozlov: I consider the flag make sense only for AOT code and not for general usage. See description: [globals.hpp#L132](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/runtime/globals.hpp#L132)

We don't want generate more complex encoding if we don't generate/use AOT code. Or compressed oops are off.
We had also recently discussed how we should disable some AOT features if AOT caching is off - the conclusion was that we should silently ignore them (may be produce UL output on demand). That is why I used ERGO setting.

src/hotspot/share/memory/memoryReserver.cpp
vnkozlov: I can do that. I got names from [CompressedOops::Mode](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/oops/compressedOops.hpp#L67) values

src/hotspot/share/cds/cdsConfig.cpp
ashu-mehra: okay

vnkozlov:
@ashu-mehra I refactored code as you asked.

vnkozlov:
@ashu-mehra I refactored code as you asked.

I looked on our old code where we added this flag and found that I missed 2 other places to check which I added now.

ashu-mehra (approved):
looks good to me

vnkozlov:
Thank you. @ashu-mehra

adinn (approved):
Looks good.

vnkozlov:
Thank you, @adinn

PR #68, opened 2025-05-16
8357060: [premain] assert(left >= right) failed: avoid underflow

ashu-mehra:
This PR fixes a few things in the premain:
1. When storing metadata in aot code cache, `AOTCacheAccess::delta_from_shared_address_base` was incorrectly using `SharedBaseAddress` for computing the offset. It should be using `MetaspaceShared::requested_base_address` because we convert the input address to the requested address.
2. Fixing the above issue results in crash in C1 and C2 compiled code during production run because CompressedKlassPointer::base() value was hardcoded in the generated code. In mainline we emit relocation for `CompressedKlassPointer::base()`. This patch adds the same changes to premain. In addition to that, it also modifies `MacroAssembler::decode_and_move_klass_not_null` and `MacroAssembler::encode_and_move_klass_not_null` which are used by C2 compiled code.
3. Fixing 2 reveals another problem when preload the code. `AOTCodeEntry::_method` can be invalid if the AOT Cache gets mapped to different address than the "requested" address, and can result in crash when accessing `AOTCodeEntry::_method` during preload. Fix is to store the offset of the `AOTCodeEntry::_method` and use the offset on load to get the correct Method pointer.
4. While working on this issue, I realized archived `AOTCodeCache::compile_nmethod` is using archived nmethod to print the assembly. This results in crash as archived nmethod has some state cleaned up. Updated `AOTCodeCache::compile_nmethod` to fix this issue.

---------
### Progress

### Issue
 * [JDK-8357060](https://bugs.openjdk.org/browse/JDK-8357060): [premain] assert(left &gt;= right) failed: avoid underflow (**Bug** - P4)

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - Committer) ⚠️ Review applies to [82567bab](https://git.openjdk.org/leyden/pull/68/files/82567bab6da5a006fbc7de44fca0264e4b433ae2)
 * [Andrew Dinn](https://openjdk.org/census#adinn) (@adinn - Committer) ⚠️ Review applies to [82567bab](https://git.openjdk.org/leyden/pull/68/files/82567bab6da5a006fbc7de44fca0264e4b433ae2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/68/head:pull/68` \
`$ git checkout pull/68`

`$ git checkout pull/68` \
`$ git pull https://git.openjdk.org/leyden.git pull/68/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 68`

`$ git pr show -t 68`

</details>

<a href="https://git.openjdk.org/leyden/pull/68.diff">https://git.openjdk.org/leyden/pull/68.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/68#issuecomment-2887030494)
</details>

### DIFF
--- a/src/hotspot/share/cds/aotCacheAccess.hpp
+++ b/src/hotspot/share/cds/aotCacheAccess.hpp
@@ -48,14 +48,31 @@ public:
   }
   static bool can_generate_aot_code(InstanceKlass* ik) NOT_CDS_RETURN_(false);
 
+  /*
+   * Used during an assembly run to compute the offset of the metadata object in the AOT Cache.
+   * The input argument is the "source" address of a metadata object (Method/Klass) loaded by the assembly JVM.
+   * Computation of the offset requires mapping the supplied metadata object to its "requested" address
+   * and subtracting that address from the requested base address.
+   * See ArchiveBuilder.hpp for definition of "source" and "requested" address.
+   */
   static uint delta_from_base_address(address addr);
 
+  /*
+   * Used during a production run to materialize a pointer to a Klass located in a loaded AOT Cache.
+   * The offset argument identifies a delta from the AOT Cache's currently mapped base address to the start of the Klass object.
+   * The offset is normally obtained by reading a value embedded in some other AOT-ed entry, like an AOT compiled code.
+   */
   static Klass* convert_offset_to_klass(uint offset_from_base_addr) {
     Metadata* metadata = (Metadata*)((address)SharedBaseAddress + offset_from_base_addr);
     assert(metadata->is_klass(), "sanity check");
     return (Klass*)metadata;
   }
 
+  /*
+   * Used during a production run to materialize a pointer to a Method located in a loaded AOT Cache.
+   * The offset argument identifies a delta from the AOT Cache's currently mapped base address to the start of the Method object.
+   * The offset is normally obtained by reading a value embedded in some other AOT-ed entry, like an AOT compiled code.
+   */
   static Method* convert_offset_to_method(uint offset_from_base_addr) {
     Metadata* metadata = (Metadata*)((address)SharedBaseAddress + offset_from_base_addr);
     assert(metadata->is_method(), "sanity check");
--- a/src/hotspot/share/cds/aotCacheAccess.hpp
+++ b/src/hotspot/share/cds/aotCacheAccess.hpp
@@ -49,7 +49,18 @@ public:
   static bool can_generate_aot_code(InstanceKlass* ik) NOT_CDS_RETURN_(false);
 
   static uint delta_from_base_address(address addr);
-  static Method* method_in_aot_code(Method* m) NOT_CDS_RETURN_(nullptr);
+
+  static Klass* convert_offset_to_klass(uint offset_from_base_addr) {
+    Metadata* metadata = (Metadata*)((address)SharedBaseAddress + offset_from_base_addr);
+    assert(metadata->is_klass(), "sanity check");
+    return (Klass*)metadata;
+  }
+
+  static Method* convert_offset_to_method(uint offset_from_base_addr) {
+    Metadata* metadata = (Metadata*)((address)SharedBaseAddress + offset_from_base_addr);
+    assert(metadata->is_method(), "sanity check");
+    return (Method*)metadata;
+  }
 
   static int get_archived_object_permanent_index(oop obj) NOT_CDS_JAVA_HEAP_RETURN_(-1);
   static oop get_archived_object(int permanent_index) NOT_CDS_JAVA_HEAP_RETURN_(nullptr);
--- a/src/hotspot/share/cds/aotCacheAccess.hpp
+++ b/src/hotspot/share/cds/aotCacheAccess.hpp
@@ -48,7 +48,7 @@ public:
   }
   static bool can_generate_aot_code(InstanceKlass* ik) NOT_CDS_RETURN_(false);
 
-  static uint delta_from_shared_address_base(address addr);
+  static uint delta_from_base_address(address addr);
   static Method* method_in_aot_code(Method* m) NOT_CDS_RETURN_(nullptr);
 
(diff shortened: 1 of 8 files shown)


ashu-mehra:
@vnkozlov @adinn can you please review these changes.

ashu-mehra:
@shipilev - fyi  after these changes we should also be able to set ArchiveRelocationMode to 1 by default in premain (See  hhttps://bugs.openjdk.org/browse/JDK-8340474).

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -1993,6 +1995,8 @@ void AOTCodeCache::preload_startup_code(TRAPS) {
       if (entry->not_entrant()) {
         continue;
       }
+      Method* m = (Method*)((address)SharedBaseAddress + entry->method_offset());
```
vnkozlov, line 1998: Why not `MetaspaceShared::requested_base_address()` here?

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -3450,7 +3454,9 @@ void AOTCodeAddressTable::init_extrs() {
   SET_ADDRESS(_extrs, SharedRuntime::handle_wrong_method);
   SET_ADDRESS(_extrs, SharedRuntime::handle_wrong_method_abstract);
   SET_ADDRESS(_extrs, SharedRuntime::handle_wrong_method_ic_miss);
-
+#if defined(AARCH64)
```
vnkozlov, line 3457: Exclude Zero.

src/hotspot/share/code/aotCodeCache.cpp
ashu-mehra: Because when loading the AOT Cache, SharedBaseAddress is set to the mapped base address (in `MetaspaceShared::map_archives`). So it is fine to use it here.

ashu-mehra: Same applies to `AOTCodeReader::read_klass` and `AOTCodeReader::read_method`.

vnkozlov (changes requested):
Can you add new methods to  `AOTCacheAccess` to get addresses based on offsets instead of: `(Method*)((address)SharedBaseAddress + entry->method_offset());` ?

ashu-mehra:
> Can you add new methods to AOTCacheAccess to get addresses based on offsets instead of: (Method*)((address)SharedBaseAddress + entry->method_offset()); ?

Yes, that make sense. I will add the method in `AOTCacheAccess`.

ashu-mehra:
@vnkozlov I added changes as per your suggestion.

vnkozlov (approved):
Good.

shipilev:
> @shipilev - fyi after these changes we should also be able to set ArchiveRelocationMode to 1 by default in premain (See https://bugs.openjdk.org/browse/JDK-8340474).

I agree. I flipped `ArchiveRelocationMode = 1` on top of this PR, and the testing started to pass. Thanks! I opened the dependent PR here: https://github.com/openjdk/leyden/pull/69

src/hotspot/share/cds/aotCacheAccess.hpp
```diff
@@ -48,8 +48,19 @@ class AOTCacheAccess : AllStatic {
   }
   static bool can_generate_aot_code(InstanceKlass* ik) NOT_CDS_RETURN_(false);
 
-  static uint delta_from_shared_address_base(address addr);
-  static Method* method_in_aot_code(Method* m) NOT_CDS_RETURN_(nullptr);
+  static uint delta_from_base_address(address addr);
```
adinn, line 51: Could we have a comment here to distinguish the operation of `delta_from_base_address` and `convert_offset_to_klass/method.

As I understand it:

`convert_offset_to_method/klass` are used during a production run to retrieve a pointer to a `Method/Klass` located in a loaded archive. The offset argument identifies a delta from the archive's currently mapped shared base address to the start of the object. It will normally have been obtained by reading a value embedded in some other archive entry.

`delta_from_base_address` are used during an assembly run to compute the offset at which a metadata object will be found in the mapped region of an archive under construction. The input argument is the address of a metadata object (`Method/Klass`) loaded by the assembly JVM. Computation of the offset requires mapping the supplied metadata object to a clean copy which resides in the current mapped archive range, remapping the address of that copy to its target location in some requested address range and subtracting that address from the requested base address (the latter being the address at which the shared archive is expected to be remapped). n.b. the offset of the clean copy in the current shared address range may not be the same as the offset of the object in the requested address range.

adinn (approved):
Looks good apart from maybe clarifying the accessor methods.

src/hotspot/share/cds/aotCacheAccess.hpp
ashu-mehra: I added comments based on your suggestion.

ashu-mehra:
@vnkozlov @adinn thanks for the reviews.

iklam:
@ashu-mehra does any part of this PR need to be upstreamed to the mainline?

vnkozlov:
> @ashu-mehra does any part of this PR need to be upstreamed to the mainline?

Compressed classed encoding changes are already in mainline as part of AOT runtime blobs caching changes.
Metadata access fix is not applicable to mainline since they are for nmethods only.

ashu-mehra:
@iklam nope, I don't think any of these issues are present in mainline.
In mainline only C1 runtime stubs are accessing `CompressedKlassPointer::base()` which is already taken care of.
I don't think any of the C2 runtime stubs access `CompressedKlassPointer::base()` which means changes to `MacroAssembler::decode_and_move_klass_not_null` and `MacroAssembler::encode_and_move_klass_not_null` shouldn't be required. Do you agree @vnkozlov @adinn ?

PR #74, opened 2025-06-03
8358343: [leyden] Drop notify_all in CompilationPolicyUtils::Queue::pop

shipilev:
Found this when reading premain-vs-mainline webrev. Mainline does not have `notify_all` in this method:
https://github.com/openjdk/jdk/blob/c382da579884c28f2765b2c6ba68c0ad4fdcb2ce/src/hotspot/share/compiler/compilationPolicy.hpp#L85-L92

But if you remove `notify_all()` in `premain`, then tests start to deadlock, see bug for a sample. The culprit is `CompilationPolicy::flush_replay_training_at_init`, which is only present in premain. I fixed it by using timed waits, which obviates the need for extra notifications. We only enter this method with `-XX:+AOTVerifyTrainingData`, so we don't care much about its performance. This is IMO better than doing a questionable `notify_all` followed by `wait` in load-bearing code.

Additional testing:
 - [x] Linux x86_64 server fastdebug, `runtime/cds` (5x, no timeouts yet; still running more iterations)

### DIFF
--- a/src/hotspot/share/compiler/compilationPolicy.cpp
+++ b/src/hotspot/share/compiler/compilationPolicy.cpp
@@ -185,7 +185,10 @@ void CompilationPolicy::replay_training_at_init_impl(InstanceKlass* klass, TRAPS
 void CompilationPolicy::flush_replay_training_at_init(TRAPS) {
    MonitorLocker locker(THREAD, TrainingReplayQueue_lock);
    while (!_training_replay_queue.is_empty_unlocked()) {
-     locker.wait(); // let the replay training thread drain the queue
+     // Let the replay training thread drain the queue.
+     // We need to re-check the queue periodically, since draining
+     // threads do not notify us about queue going empty.
+     locker.wait(100);
    }
 }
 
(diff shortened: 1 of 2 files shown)


shipilev:
@veresov @iwanowww ^

veresov:
Is there a reason why can't we just do the processing work in the thread calling the flush instead of waiting for the replay thread? That is, why not make it be like this:

```
void CompilationPolicy::flush_replay_training_at_init(TRAPS) {
  InstaceKlass* ik;
  do {
    ik = _training_replay_queue.try_pop(TrainingReplayQueue_lock, THREAD);
    if (ik != nullptr) {
      replay_training_at_init_impl(ik, THREAD);
    }
  } while (ik != nullptr);
}
```

shipilev:
> Is there a reason why can't we just do the processing work in the thread calling the flush instead of waiting for the replay thread? That is, why not make it be like this:

You tell me :) I guess one upside of current code is to leave draining/processing in one place/thread, and thus never run into false positives/negatives due to diagnostic code (this hunk, gated by `-XX:+AOTVerifyTrainingData`) doing something that production code does not. So I mildly prefer it in current form. I can change it, if you want.

veresov:
Actually the current approach (even with the spin-wait) and my solution too are not really correct. The fact that the queue is empty doesn't mean that every last item has been processed. The last item may have been popped, but still is being worked on. So however you look at it we should set up some kind of a handshake to make sure the replay thread is done processing, not just done popping.

PR #78, opened 2025-06-06
8358546: [premain] AOT code generation should support UseCompactObjectHeaders

vnkozlov:
COH testing was added to tier3 which we use in premain branch testing. And we noticed that `runtime/cds/appcds/aotCode/` tests crashes during "production" run when UseCompactObjectHeaders is on.

There were five issues when AOT code is generated with COH:
 - missing CompressedKlassPointers::base() relocation on x86 in `MacroAssembler::decode_and_move_klass_not_null()`
 - missing CKP:base() relocation on aarch64 in `MacroAssembler::cmp_klass()` because `decode_klass_not_null_for_aot()` was not called
 - On aarch64 full AOT encoding/decoding code for compressed klass pointers was generated even when base is NULL. Do special AOT code only when base is not NULL, similar to code on x64.
 - Record CKP::basse in AOT code cache configuration and check it on load similar to what we do for compressed oops base.
 - Prototype header value for new allocated object is embedded into C2 generated code as constant. But with COH the prototype also contains encoded class. This class encoding (and prototype header) could be change when CDS store this information in archive. We need to load it in code.

Tested tier1-3

---------
### Progress

### Issue
 * [JDK-8358546](https://bugs.openjdk.org/browse/JDK-8358546): [premain] AOT code generation should support UseCompactObjectHeaders (**Bug** - P4)

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/78/head:pull/78` \
`$ git checkout pull/78`

`$ git checkout pull/78` \
`$ git pull https://git.openjdk.org/leyden.git pull/78/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 78`

`$ git pr show -t 78`

</details>

<a href="https://git.openjdk.org/leyden/pull/78.diff">https://git.openjdk.org/leyden/pull/78.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/78#issuecomment-2948104540)
</details>

### DIFF
--- a/src/hotspot/share/code/aotCodeCache.cpp
+++ b/src/hotspot/share/code/aotCodeCache.cpp
@@ -770,13 +770,13 @@ bool AOTCodeCache::Config::verify() const {
 
   if ((_compressedKlassBase == nullptr || CompressedKlassPointers::base() == nullptr) && (_compressedKlassBase != CompressedKlassPointers::base())) {
     log_debug(aot, codecache, init)("AOT Code Cache disabled: incompatible CompressedKlassPointers::base(): %p vs current %p", _compressedKlassBase, CompressedKlassPointers::base());
-    AOTStubCaching = false;
+    return false;
   }
 
   // This should be the last check as it only disables AOTStubCaching
   if ((_compressedOopBase == nullptr || CompressedOops::base() == nullptr) && (_compressedOopBase != CompressedOops::base())) {
     log_debug(aot, codecache, init)("AOTStubCaching is disabled: incompatible CompressedOops::base(): %p vs current %p", _compressedOopBase, CompressedOops::base());
-    AOTStubCaching = false;
+    return false;
   }
 
   return true;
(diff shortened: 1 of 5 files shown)


src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -767,6 +768,11 @@ bool AOTCodeCache::Config::verify() const {
     return false;
   }
 
+  if ((_compressedKlassBase == nullptr || CompressedKlassPointers::base() == nullptr) && (_compressedKlassBase != CompressedKlassPointers::base())) {
```
shipilev, line 771: I am guessing we only disable `AOTStubCaching`, and not `AOTCodeCaching` wholesale, because the normal code deals with this properly. Correct?

shipilev (approved):
Looks reasonable.

src/hotspot/share/code/aotCodeCache.cpp
vnkozlov: yes
I still did not figure out what is wrong with stubs.

vnkozlov:
Thank you for review Aleksey

src/hotspot/share/code/aotCodeCache.cpp
vnkozlov: Hmm, actually no. Normal JIT code will be also affected except adapters I think. This code from mainline and have to be adjusted for premain.

vnkozlov: I decide just `return false` for all cases here. I will adjust it in other changes.

PR #80, opened 2025-06-19
Cleanup unused _decompile and _ignore_decompile fields in AOTCodeEntry

ashu-mehra:
New workflow always sets `_ignore_decompile` to true, which implies `AOTCodeEntry::_decompile` is not in `check_entry` and is thus redundant. These two fields can just be removed.


vnkozlov (commented):
Good. I will test it.

ashu-mehra:
@vnkozlov thanks for the review and testing it.

PR #81, opened 2025-06-24
8350550: Preload classes from AOT cache during VM bootstrap

iklam:
This is still work-in-progress

---------
### Progress
- [ ] Change must not contain extraneous whitespace

### Issue
 * [JDK-8350550](https://bugs.openjdk.org/browse/JDK-8350550): Preload classes from AOT cache during VM bootstrap (**Enhancement** - P3)

### Reviewers
 * [John R Rose](https://openjdk.org/census#jrose) (@rose00 - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/81/head:pull/81` \
`$ git checkout pull/81`

`$ git checkout pull/81` \
`$ git pull https://git.openjdk.org/leyden.git pull/81/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 81`

`$ git pr show -t 81`

</details>

<a href="https://git.openjdk.org/leyden/pull/81.diff">https://git.openjdk.org/leyden/pull/81.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/cds/aotClassLinker.cpp
+++ b/src/hotspot/share/cds/aotClassLinker.cpp
@@ -196,7 +196,7 @@ void AOTClassLinker::write_to_archive() {
 
   if (CDSConfig::is_dumping_aot_linked_classes()) {
     AOTLinkedClassTable* table = AOTLinkedClassTable::get(CDSConfig::is_dumping_static_archive());
-    table->set_boot(write_classes(nullptr, true));
+    table->set_boot1(write_classes(nullptr, true));
     table->set_boot2(write_classes(nullptr, false));
     table->set_platform(write_classes(SystemDictionary::java_platform_loader(), false));
     table->set_app(write_classes(SystemDictionary::java_system_loader(), false));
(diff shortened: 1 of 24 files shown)


rose00 (approved):
Very nice work.  This is a Big Transaction that inhales the initial set of classes.  I see it scales all the way up to the app class loader.  Very cool, ship it!

Some documentation-related comments follow:

I noticed patching of the relation of class mirror to module.  This is present even if the full module graph is being used.  Wouldn’t it be possible to pre-configure that in the AOT cache, and only do the patching if there is a non-default module graph?  Or am I misreading the code?  (I’m looking at restore_module_of_preloaded_classes etc.)

The restore_unshareable_info call in SystemDictionary::preload_class is used for other miscellanous links from the class, similar to the module links I just mentioned.  Same question, I guess:  Why are these not pre-materialized in the AOT cache?  They are “unshareable” for a reason; where is that reason documented?

And then another question:  Why not handle the module linkage at the same place as the package linkage and the PD linkage?  Possible answer: The module graph gets created last, so it must be back-patched.  Where is that fact documented?

PR #84, opened 2025-07-04
Store cpu features in AOTCodeCache header

ashu-mehra:
This is the initial version of storing cpu features in the AOTCodeCache to verify runtime env has the same cpu capabilities as the assembly env. It covers both x86 and aarch64.
AOTCodeCache header is updated to store the cpu features in arch-dependent form (although its same for currently supported architectures - x86 and aarch64).

~It also fixes a bug - the `polling_page_vectors_safepoint_handler_blob` can be null if AVX is not present on a system. This causes crash as this blob's entry point is stored in the address table.
I came across this when I did the assembly run with -XX:UseAVX=0 option.~

---------
### Progress

### Reviewers
 * [Andrew Dinn](https://openjdk.org/census#adinn) (@adinn - Committer) ⚠️ Review applies to [fff8cd11](https://git.openjdk.org/leyden/pull/84/files/fff8cd116b97c9fe3477010c43a2da8580a39d0b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/84/head:pull/84` \
`$ git checkout pull/84`

`$ git checkout pull/84` \
`$ git pull https://git.openjdk.org/leyden.git pull/84/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 84`

`$ git pr show -t 84`

</details>

<a href="https://git.openjdk.org/leyden/pull/84.diff">https://git.openjdk.org/leyden/pull/84.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/84#issuecomment-3037207641)
</details>


ashu-mehra:
@vnkozlov @adinn please take a look

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -725,3 +732,67 @@ void VM_Version::initialize_cpu_information(void) {
...
+  return sizeof(_features);
+}
+
+void VM_Version::store_cpu_features(void* buf) {
+  *(uint64_t*)buf = _features;
+}
+
+bool VM_Version::supports_features(void* features_buffer) {
+  uint64_t features_to_test = *(uint64_t*)features_buffer;
+  return (_features & features_to_test) == features_to_test;
```
jankratochvil, line 773: Here you test whether the new machine has at least the same CPU features as the old (stored) machine. From our experience from [CRaC](https://docs.azul.com/core/crac/cpu-features) one should rather compare the features are equal. As some features presence affect ABI of the JIT-compiled code and if the new machine has such CPU features the newly JITted code will be incompatible with ABI of the stored code cache.

adinn: Could you provide more detail about your statement "some features presence affect ABI of the JIT-compiled code"?

jankratochvil: Simply said if you store the compiled Java code on old CPU (almost no features) and then run it on new CPU (full features) then it will crash. It crashes when a newly compiled code calls the restored code or vice versa. If you disagree I can try to reproduce with Leyden, it was happening with CRaC.

ashu-mehra: @jankratochvil thanks for sharing your experience. It would be great if you can provide a reproducer.

jankratochvil:
It can be a future extension but Leyden may also find useful an equivalent of `-XX:CPUFeatures=generic` so that you can store the data on any machine and restore it also on any machine - limiting the CPU feature set in all runs to some reasonable minimum.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
adinn: @jankratochvil I'm not disagreeing - merely wanting to see where CRaC goes wrong and why (noting that it is a slightly different circumstance to the AOT Cache case).

My preferred position would be that there /ought not/ to be any issues reusing code generated during the AOT Cache assembly phase in the absence of some given feature in combination with with code generated in a production run where that feature is present i.e. that that any incompatibility would constitute an error in the code generation scheme. I'd be very interested in any evidence you can provide to show that 1) such a code generation error exists or, worse, 2) my preference is unrealizable.

adinn:
> It can be a future extension but Leyden may also find useful an equivalent of -XX:CPUFeatures=generic so that you can store the data on any machine and restore it also on any machine - limiting the CPU feature set in all runs to some reasonable minimum.
We have discussed such an option in dev team meetings. We are still considering what would be the most useful and/or usable model.

src/hotspot/cpu/x86/vm_version_x86.hpp
```diff
@@ -366,83 +366,83 @@ class VM_Version : public Abstract_VM_Version {
...
 #define CPU_FEATURE_FLAGS(decl) \
-    decl(CX8,               "cx8",               0)  /*  next bits are from cpuid 1 (EDX) */ \
-    decl(CMOV,              "cmov",              1)  \
-    decl(FXSR,              "fxsr",              2)  \
-    decl(HT,                "ht",                3)  \
+    decl(CX8,               cx8,               0)  /*  next bits are from cpuid 1 (EDX) */ \
+    decl(CMOV,              cmov,              1)  \
+    decl(FXSR,              fxsr,              2)  \
+    decl(HT,                ht,                3)  \
                                                      \
```
vnkozlov, line 375: Please, revert alignment change to reduce size of changes. We can do it in mainline later.

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -3227,7 +3249,10 @@ void AOTCodeAddressTable::init_shared_blobs() {
   SET_ADDRESS(_shared_blobs, SharedRuntime::polling_page_safepoint_handler_blob()->entry_point());
   SET_ADDRESS(_shared_blobs, SharedRuntime::polling_page_return_handler_blob()->entry_point());
 #ifdef COMPILER2
-  SET_ADDRESS(_shared_blobs, SharedRuntime::polling_page_vectors_safepoint_handler_blob()->entry_point());
+  // polling_page_vectors_safepoint_handler_blob can be nullptr if AVX feature is not present or is disabled
+  if (SharedRuntime::polling_page_vectors_safepoint_handler_blob() != nullptr) {
```
vnkozlov: I will push this separately.

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -649,6 +649,12 @@ void AOTCodeCache::Config::record() {
   _contendedPaddingWidth = ContendedPaddingWidth;
   _objectAlignment       = ObjectAlignmentInBytes;
   _gc                    = (uint)Universe::heap()->kind();
+  _cpu_features_size     = (uint)VM_Version::cpu_features_size();
+  void* cpu_features_buffer = (void*)(this+1);
```
vnkozlov: I don't like. Store it in separate space and reference (offset in archive) it from Config.

src/hotspot/share/code/aotCodeCache.hpp
```diff
@@ -269,6 +270,7 @@ class AOTCodeAddressTable : public CHeapObj<mtCode> {
 public:
   AOTCodeAddressTable() :
     _extrs_addr(nullptr),
+    _stubs_addr(nullptr),
```
vnkozlov: I pushed it already into premain.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
rvansa: I can't find the logs from when I was investigating the issue, but AFAIR https://github.com/openjdk/crac/pull/103 was motivated by a bug that happened in compiler thread; it was going through some code that calculated buffer size for output code based on the availability of CPU features, and then it went to actually write down the instructions. When the checkpoint happened in the middle of this and the CPU got changed (we got a 'better' CPU) the decision in this codepath was changed, and resulted in a buffer overrun.
So it was rather a synchronization problem: some code was written assuming that the CPU features are runtime-constant, but these are not. There is certainly space for a better solution, but we would have to track through some complex code and make sure that it works on a 'snapshot' of features.

ashu-mehra: @rvansa 

>  it was going through some code that calculated buffer size for output code based on the availability of CPU features, and then it went to actually write down the instructions. When the checkpoint happened in the middle of this and the CPU got changed (we got a 'better' CPU) the decision in this codepath was changed, and resulted in a buffer overrun.

I can imagine this happening in the context of checkpoint-snapshot but I don't think think AOTCodeCache can hit this issue of buffer overrun. Code generation is not suspended-resumed in Leyden workflow. When the code generation starts, it is always completed before the JVM exits the assembly phase.

> So it was rather a synchronization problem: some code was written assuming that the CPU features are runtime-constant, but these are not. There is certainly space for a better solution, but we would have to track through some complex code and make sure that it works on a 'snapshot' of features.

Other than this buffer overrun problem, have you come across any other code that relies on the assumption that CPU features are runtime-constant?

src/hotspot/cpu/x86/vm_version_x86.hpp
ashu-mehra: This change is not about alignment, but removing the quotes around the cpu feature names and use XSTR macro in DECLARE_CPU_FEATURE_NAME macros instead. Any change in alignment is the consequence of that.
Do you want me to go back to using quotes and remove XSTR usage?

src/hotspot/share/code/aotCodeCache.cpp
ashu-mehra: okay,  I will take this out and push it separately.

src/hotspot/cpu/x86/vm_version_x86.hpp
vnkozlov: Okay, I missed that. Keep XSTR. Why aarch64 code does not use XSTR?

ashu-mehra:
@vnkozlov I have pushed changes addressing your review comments. Please review again.

src/hotspot/share/code/aotCodeCache.cpp
ashu-mehra: I misunderstood this. I see you have already pushed it. Thanks for that!

vnkozlov:
@ashu-mehra you have to remove that (UseSSE/UseAVX) code in your changes - they will be negated anyway. Yes, it was more flexible to allow old AOT code run on new machines. But we can do that with your changes later.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -725,3 +732,67 @@ void VM_Version::initialize_cpu_information(void) {
 
   _initialized = true;
 }
+
+void VM_Version::insert_features_names(uint64_t features, char* names_buf, size_t buf_size) {
```
rose00: Uh, there is a DRY failure here, which leads to a bug.  (DRY = Don't Repeat Yourself.)

It appears to be a modified copy from another port, and this copy has a buffer overflow bug in the modifications.

And the same bug occurs in another modified copy.

Reviewers and porters should not have to repeatedly debug the same subtle code in multiple platform-specific files.

The bug is that `buf_rem` is a dead variable, while loop-invariant `buf_size` is passed with `buf` to `snprintf`.  A hasty reader might think that `buf` and `buf_len` vary together but they don't; this is a bad naming choice that masks the bug.

Instead of point-fixing the bug (in at least two places), please consider factoring this out somewhere that all platforms can use it, and then all the reviewers can make sure that the single copy is bug-free.  Otherwise, we have a new bug surface on all future ports.  (Yes, this a small thing, but let's do it right.)

I suggest making this be a template function that takes a closure (of type `int->char*`) and iterates it over a range `0..N-1`, accumulating the closure-produced strings (if not null) via `snprintf` into a sized buffer, safely.  Return the number of written characters.  I looked for possibly related APIs and found FormatBuffer but it is poorly documented.  I suggest just making a public static helper function in class outputStream.  That's clean enough IMO.

```
public: static
template<typename FN>
size_t insert_string_list(char* buf, size_t buf_size, int start, int limit, FN fn) {
  size_t np = 0;
  ...
  for (int i = start; i < limit; i++) {
    const char* str = fn(i);
    if (str == nullptr)  continue;
    const char* comma = (first && !use_first_comma) ? "" : ", ";
    auto res = jio_snprintf(buf + np, buf_size - np, "%s%s", comma, str);
    np += res;
    if (res < 0 || np+1 > buf_size)  return -1;
  }
  return np;
}
```

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
rvansa: @ashu-mehra I don't recall that, but don't have too much personal experience switching the CPU between runs.

One issue that could do trouble is the hyperthreading support; I have one of those hybrid Intels with performance and efficiency cores. HT flag is on on the performance cores, and when you record the features on perf core and try to restore on Atom, a straightforward bitwise comparison would error, even though HT is not something the generated code usually relies on.

theRealAph: > there /ought not/ to be any issues reusing code generated during the AOT Cache assembly phase in the absence of some given feature in combination with with code generated in a production run where that feature is present i.e. that that any incompatibility would constitute an error in the code generation scheme.

Yes. In fact, I'd call any failure to use a feature "a bug" that we ought to look at. At best it's a code smell. As one of Arthur C Clarke's editors once put it, CH3CH2SH.

theRealAph: Sorry, that was badly worded. It's never _compulsory_ to use any feature, and if any code makes the  assumption that if a feature is supported and turned on then it must be used, then that code has a bug.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
ashu-mehra: @rose00 Thanks for catching that bug. I have refactored the code a bit more than you suggested by adding the static method in the FormatBuffer. I think FormatBuffer is a good fit for this use case as the client code doesn't need to worry about the temporary buffer and size anymore. Please check https://github.com/openjdk/leyden/pull/84/commits/9018729b6cc618a38a9cc8e0ec595664e2f35e5e to see the changes in isolation from rest of the patch. Does it look good?

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -3213,10 +3246,7 @@ void AOTCodeAddressTable::init_shared_blobs() {
   SET_ADDRESS(_shared_blobs, SharedRuntime::polling_page_safepoint_handler_blob()->entry_point());
   SET_ADDRESS(_shared_blobs, SharedRuntime::polling_page_return_handler_blob()->entry_point());
 #ifdef COMPILER2
-  // polling_page_vectors_safepoint_handler_blob can be nullptr if AVX feature is not present or is disabled
-  if (SharedRuntime::polling_page_vectors_safepoint_handler_blob() != nullptr) {
-    SET_ADDRESS(_shared_blobs, SharedRuntime::polling_page_vectors_safepoint_handler_blob()->entry_point());
-  }
+  SET_ADDRESS(_shared_blobs, SharedRuntime::polling_page_vectors_safepoint_handler_blob()->entry_point());
```
vnkozlov: Mistake update?

vnkozlov (commented):
`src/hotspot/share/utilities/formatBuffer.hpp` changes should be done separately in mainline. And then ported into premain branch.

src/hotspot/share/code/aotCodeCache.cpp
ashu-mehra: oh! yes, that's a mistake. I will remove it.

vnkozlov:
> should I remove the patch that introduced these changes and revert back to earlier implementation?

It is fine to keep it since it will go into mainline.

vnkozlov:
@ashu-mehra 

Look on `test/hotspot/jtreg/compiler/arraycopy/stress/TestStressArrayCopy.java` how to get CPU features using WB instead of running separate processes.

Also add ` * @requires os.simpleArch == "x64"` if you want to run it only on x64.

vnkozlov:
I submitted leyden/premain testing for current 04 version.

vnkozlov (approved):
This looks good. My testing "passed" - I don't anything which could be caused by these changes.

Please, before pushing it run performance testing with CPU features check and without it (you have flag now) to see if we hit bailouts (no AOT code usage) due to some bug.

Thanks.

vnkozlov (approved):
Good.

vnkozlov:
Yes, thank you for testing that.

vnkozlov:
> I see some failures on macos-aarch64 in GHA testing:
> 
> * tier1:  `jdk/java/lang/Thread/virtual/stress/PingPong.java` and `java/lang/Thread/virtual/stress/Skynet.java#default` failing with time outs on macos-aarch64
> * tier2: `java/util/concurrent/BlockingQueue/ProducerConsumerLoops.java`
> * tier3: `java/foreign/TestConcurrentClose.java`
> 
> They all fail due to timeouts. Has anyone seen these failures before? I don't think these can be related to my changes.

I see timeout only for this test.in latest testing of premain:
   - runtime/Monitor/ConcurrentDeflation.java

My latest testing of your changes showed timeouts:
 - tier1 the same as you see and java/util/concurrent/BlockingQueue/SingleProducerMultipleConsumerLoops.java
 - tier3: several serviceability/jvmti/stress/StackTrace tests

vnkozlov:
Let me test your changes again with latest premain.

ashu-mehra:
> However it assumes a lot of things fromthe (current) fact that this must be Apple silicon so things do not match that closely.

I didn't follow your comment. Did you mean there is code that applies to Apple silicon but not to aarch64?

> Interesting that none of the macos gate tests failed even though you changed nothing in this file . . .

Yeah I am surprised I didn't get any failures to indicate the missing changes.

adinn:
@ashu-mehra I was merely indicating that the changes you made in vm_version_linux_aarch64.cpp are not simply able to be reapplied to vm_version_bsd_aarch64.cpp because the latter file checks things very differently.

In particular, it bypasses a lot of the hardware checks on the assumption that this must be apple hardware (perhaps a questionable thing to rely on).

ashu-mehra:
It seems vm_version_bsd_aarch64.cpp changes fixed the macos-aarch64 timeout failures. No test failed this time.

vnkozlov:
> @adinn @vnkozlov does this look good to go?

May be we should wait when #26515 is finalized to use in these changes.

Meanwhile I can re-submit testing with latest changes and see if timeouts indeed gone.

vnkozlov:
My testing passed - no unusual timeouts now. Good.

adinn (approved):
Looking good.

ashu-mehra:
@vnkozlov @adinn  I pulled the changes for JDK-8364128 into this PR and merged the premain changes. Do you want to give this another look?

vnkozlov:
What are GHA testing failures?

ashu-mehra:
my tier1 testing passed. Only failures were the gtest based ones because I didn't use --with-gtest.

vnkozlov (approved):
My testing passed.

ashu-mehra:
thanks @vnkozlov @adinn for the review and testing.

PR #85, opened 2025-07-17
8362534: [Leyden] assert(comp != nullptr) failed: Ensure we have a compiler

ashu-mehra:
Precompiler tries to compile methods at all levels without taking into account TieredStopAtLevel value.
This patch updates Precompiler to restrict compilation to only levels allowed by TieredStopAtLevel.


ashu-mehra:
@vnkozlov can you please review.

vnkozlov (approved):
Looks good. You can push it.

PR #86, opened 2025-07-18
8362657: Make tables used in AOT assembly phase GC-safe

iklam:
This fixes crashes when using `-Xlog:aot+map=trace,aot+map+oops=trace:file=aot.oops.txt:none:filesize=0` during the assembly phase.

I basically changed a few raw `oop` pointers to `OopHandle`. I also simplified `ArchiveHeapWriter::is_marked_as_native_pointer()`.

Note that for `HeapShared::archived_object_cache()`, I calculate the hashcode using the raw oop. This way, we can avoid calling `oopDesc::identity_hash()` for every archived oops. As a result, when we use this table after exiting the CDS safepoint, we must rehash it. See changes in `ArchiveBuilder::CDSMapLogger::log()`.

- It might be alright to pre-calculate the identity hash of the archived oops in the assembly phase, but I think we should do that only if we find a reason to do so. We shouldn't do it just because it's convenient for implementing an internal table,

---------
### Progress

### Issue
 * [JDK-8362657](https://bugs.openjdk.org/browse/JDK-8362657): Make tables used in AOT assembly phase GC-safe (**Enhancement** - P4)

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - Committer) ⚠️ Review applies to [123b766c](https://git.openjdk.org/leyden/pull/86/files/123b766c21e2a19b29633886cdf486d037f9230f)
 * [Ashutosh Mehra](https://openjdk.org/census#asmehra) (@ashu-mehra - Committer) ⚠️ Review applies to [123b766c](https://git.openjdk.org/leyden/pull/86/files/123b766c21e2a19b29633886cdf486d037f9230f)
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/86/head:pull/86` \
`$ git checkout pull/86`

`$ git checkout pull/86` \
`$ git pull https://git.openjdk.org/leyden.git pull/86/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 86`

`$ git pr show -t 86`

</details>

<a href="https://git.openjdk.org/leyden/pull/86.diff">https://git.openjdk.org/leyden/pull/86.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/86#issuecomment-3090777803)
</details>

### DIFF
--- a/src/hotspot/share/cds/heapShared.hpp
+++ b/src/hotspot/share/cds/heapShared.hpp
@@ -427,7 +427,7 @@ private:
   static int append_root(oop obj);
   static GrowableArrayCHeap<OopHandle, mtClassShared>* pending_roots() { return _pending_roots; }
 
-  static void delete_tables_with_raw_oops() NOT_CDS_JAVA_HEAP_RETURN; 
+  static void delete_tables_with_raw_oops() NOT_CDS_JAVA_HEAP_RETURN;
 
   // Dump-time and runtime
   static objArrayOop root_segment(int segment_idx);
--- a/src/hotspot/share/cds/heapShared.hpp
+++ b/src/hotspot/share/cds/heapShared.hpp
@@ -384,7 +384,6 @@ private:
   static ArchivedObjectCache* archived_object_cache() {
     return _archived_object_cache;
   }
-  static void rehash_archived_object_cache();
 
   static CachedOopInfo* get_cached_oop_info(oop orig_obj) {
     OopHandle oh(&orig_obj);
@@ -427,8 +426,6 @@ private:
   static int append_root(oop obj);
   static GrowableArrayCHeap<OopHandle, mtClassShared>* pending_roots() { return _pending_roots; }
 
-  static void delete_tables_with_raw_oops() NOT_CDS_JAVA_HEAP_RETURN;
-
   // Dump-time and runtime
   static objArrayOop root_segment(int segment_idx);
   static oop get_root(int index, bool clear=false);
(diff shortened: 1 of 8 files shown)


vnkozlov (approved):
Seems fine. Need second review.

ashu-mehra:
@iklam there are arm compile failures likely due to some missing NOT_CDS_ macros. Otherwise this looks good to me.

ashu-mehra (approved):

src/hotspot/share/cds/heapShared.hpp
```diff
@@ -382,6 +385,11 @@ class HeapShared: AllStatic {
     return _archived_object_cache;
   }
 
+  static CachedOopInfo* get_cached_oop_info(oop orig_obj) {
+    OopHandle oh(&orig_obj);
```
shipilev, line 389: `OopHandle`-s have no destructors, so this leaks?

iklam: This only allocates a temporary `OopHandle` on the stack for the purpose of doing a look-up. The storage of this is allocated on the stack (the address of the local variable `obj`). So there's no leak here.

```
  OopHandle oh(&obj);
  return archived_object_cache()->get(oh) != nullptr;
```

When an `OopHandle` is stored in the `archived_object_cache()`, we will allocate the storage from `Universe::vm_global()`. This  storage should be freed when the `archived_object_cache()` is deleted, but we currently never delete that table, as we assume the VM will soon exit after AOT assembly is done.

shipilev: Oh, there is a special constructor, I missed that:
https://github.com/openjdk/jdk/blob/ed70910b0f3e1b19d915ec13ac3434407d01bc5d/src/hotspot/share/oops/oopHandle.hpp#L47C1-L47C42

I thought we go through a normal allocation path that leaks the slots in relevant `OopStorage`.

shipilev (approved):
OK, this looks fine.

shipilev (approved):

iklam:
Thanks @vnkozlov @ashu-mehra @shipilev for the review
/integrate

PR #87, opened 2025-07-21
8362558: [Leyden] fatal error: New workflow: should not compile code for unarchived class: UnsupportedBSMs$$Lambda/0...

iklam:
We should not AOT-compile indy call sites that cannot be archived in the resolved state.

### DIFF
--- a/src/hotspot/share/ci/ciEnv.cpp
+++ b/src/hotspot/share/ci/ciEnv.cpp
@@ -22,6 +22,7 @@
  *
  */
 
+#include "cds/aotConstantPoolResolver.hpp"
 #include "cds/archiveBuilder.hpp"
 #include "cds/cdsConfig.hpp"
 #include "ci/ciConstant.hpp"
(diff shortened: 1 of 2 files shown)


src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -832,6 +833,14 @@ ciMethod* ciEnv::get_method_by_index_impl(const constantPoolHandle& cpool,
     // As with other two-component call sites, both values must be independently verified.
     assert(index < cpool->cache()->resolved_indy_entries_length(), "impossible");
     Method* adapter = cpool->resolved_indy_entry_at(index)->method();
+    if (is_precompiled()) {
+        ResolvedIndyEntry* indy_info = cpool()->resolved_indy_entry_at(index);
```
vnkozlov: Can move `indy_info` up and use it in `adapter = ` expression too?

Do you need to use `()` for `cool()`? Other code does not use it.

vnkozlov:
Minimal VM build failed: I think you missed #if INCLUDE_CDS

src/hotspot/share/ci/ciEnv.cpp
iklam: I changed `cpool()->` to `cpool->`.

Do you mean this:

```
if (!AOTConstantPoolResolver::is_resolution_deterministic(cpool(),
         cpool->resolved_indy_entry_at(index)->constant_pool_index())) { 
    adapter = nullptr;
}
```

I think that will make the code hard to read as you see two "index" in a row.

vnkozlov: I ment to move up not down:
```
   ResolvedIndyEntry* indy_info = cpool->resolved_indy_entry_at(index);
   Method* adapter = indy_info->method();
#if INCLUDE_CDS
    if (is_precompiled() && !AOTConstantPoolResolver::is_resolution_deterministic(cpool(), indy_info->constant_pool_index())) {
```

iklam: But `indy_info` is not used outside of that scope, so we shouldn't move it up.

vnkozlov: What expression at the line #835 uses? Is it different?:
```
  Method* adapter = cpool->resolved_indy_entry_at(index)->method();
```

iklam: Oops, sorry I misunderstood. I updated the code as you suggested.

PR #90, opened 2025-08-07
Restore Method pointer for every AOTCodeEntry

ashu-mehra:
There was a recent commit [0] that included code to invalidate preload entry when invalidating a normal `AOTCodeEntry`. This is done by reaching out to the preload entry through the normal `AOTCodeEntry::_method`:

https://github.com/openjdk/leyden/blob/7b7648a4c9f67be509c6fccbcbc0502648388fdc/src/hotspot/share/code/aotCodeCache.cpp#L1056-L1070

But unfortunately `AOTCodeEntry::_method` is restored only for the entries marked for preload:
https://github.com/openjdk/leyden/blob/7b7648a4c9f67be509c6fccbcbc0502648388fdc/src/hotspot/share/code/aotCodeCache.cpp#L1938-L1939

This PR fixes this bug by restoring `AOTCodeEntry::_method` for all AOTCodeEntry-s. This is achieved by using AOTCache's pointer bitmap to track `AOTCodeEntry::_method`. It removes the need to store method offset separately in the AOTCodeEntry.
It also fixes a couple of related bugs:
1. In `AOTCodeCache::finish_write()` it is possible that the AOTCodeEntry array is not properly aligned.
2. When invalidating a preload entry, it is possible that the entry has not been loaded. This triggers the assert that expects an entry to be invalidated is always loaded.

[0] https://github.com/openjdk/leyden/commit/392fbbb1859cd71521cb915b601a65cf59ba495b

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - Committer) ⚠️ Review applies to [8d01aebf](https://git.openjdk.org/leyden/pull/90/files/8d01aebfd25bf847658a9f39be1c05f40a0afad3)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/90/head:pull/90` \
`$ git checkout pull/90`

Update a local copy of the PR: \
`$ git checkout pull/90` \
`$ git pull https://git.openjdk.org/leyden.git pull/90/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 90`

View PR using the GUI difftool: \
`$ git pr show -t 90`

</details>
[...]

PR #91, opened 2025-08-13
Use method offset as the hash for storing nmethods

ashu-mehra:
Currently AOTCodeEntry for nmethod uses a hash obtained from the method holder's name, method name and signature.   This is not very reliable.  With https://github.com/openjdk/leyden/pull/90, AOTCodeEntry is always linked to its corresponding Method, so we should be able to use the Method pointer as the hash for AOTCodeEntries for nmethod. However, the problem with pointers are not stable across runs. So instead of the pointer, we can use the offset of the Method in AOTCache. This should work because the Method is guaranteed to be in AOTCache.
This PR updates the code to use Method offset as the hash for AOTCodeEntries for nmethods.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/91/head:pull/91` \
`$ git checkout pull/91`

`$ git checkout pull/91` \
`$ git pull https://git.openjdk.org/leyden.git pull/91/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 91`

`$ git pr show -t 91`

</details>

<a href="https://git.openjdk.org/leyden/pull/91.diff">https://git.openjdk.org/leyden/pull/91.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/91#issuecomment-3186026186)
</details>

### DIFF
--- a/src/hotspot/share/code/aotCodeCache.cpp
+++ b/src/hotspot/share/code/aotCodeCache.cpp
@@ -903,6 +903,9 @@ uint AOTCodeCache::write_bytes(const void* buffer, uint nbytes) {
 
 AOTCodeEntry* AOTCodeCache::find_code_entry(const methodHandle& method, uint comp_level) {
   assert(is_using_code(), "AOT code caching should be enabled");
+  if (!method->is_shared()) {
+    return nullptr;
+  }
   switch (comp_level) {
     case CompLevel_simple:
       if ((DisableAOTCode & (1 << 0)) != 0) {
@@ -924,15 +927,21 @@ AOTCodeEntry* AOTCodeCache::find_code_entry(const methodHandle& method, uint com
   }
   TraceTime t1("Total time to find AOT code", &_t_totalFind, enable_timers(), false);
   if (is_on() && _cache->cache_buffer() != nullptr) {
-    ResourceMark rm;
-    const char* target_name = method->name_and_sig_as_C_string();
-    uint hash = (uint)pointer_delta((address)method(), (address)SharedBaseAddress, 1);
-    AOTCodeEntry* entry = _cache->find_entry(AOTCodeEntry::Code, hash, comp_level);
+    uint id = AOTCacheAccess::convert_method_to_offset(method());
+    AOTCodeEntry* entry = _cache->find_entry(AOTCodeEntry::Code, id, comp_level);
     if (entry == nullptr) {
-      log_info(aot, codecache, nmethod)("Missing entry for '%s' (comp_level %d, hash: " UINT32_FORMAT_X_0 ")", target_name, (uint)comp_level, hash);
+      LogStreamHandle(Info, aot, codecache, nmethod) log;
+      if (log.is_enabled()) {
+        ResourceMark rm;
+        const char* target_name = method->name_and_sig_as_C_string();
+        log.print("Missing entry for '%s' (comp_level %d, id: " UINT32_FORMAT_X_0 ")", target_name, (uint)comp_level, id);
+      }
 #ifdef ASSERT
-    } else if (method() != entry->method()) {
-      assert(false, "AOTCodeCache: saved nmethod's method %p (hash: " UINT32_FORMAT_X_0 ") is different from the method %p being looked up" , entry->method(), hash, method());
+    } else {
+      ResourceMark rm;
+      assert(method() == entry->method(), "AOTCodeCache: saved nmethod's method %p (name: %s id: " UINT32_FORMAT_X_0
+             ") is different from the method %p (name: %s, id: " UINT32_FORMAT_X_0 " being looked up" ,
+             entry->method(), entry->method()->name_and_sig_as_C_string(), entry->id(), method(), method()->name_and_sig_as_C_string(), id);
 #endif
     }
 
@@ -1644,7 +1653,7 @@ AOTCodeEntry* AOTCodeCache::write_nmethod(nmethod* nm, bool for_preload) {
   // Write name
   uint name_offset = 0;
   uint name_size   = 0;
-  uint hash = 0;
+  uint id = 0;
   uint n;
   {
     ResourceMark rm;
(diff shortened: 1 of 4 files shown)


src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -926,18 +926,13 @@ AOTCodeEntry* AOTCodeCache::find_code_entry(const methodHandle& method, uint com
...
       log_info(aot, codecache, nmethod)("Missing entry for '%s' (comp_level %d, hash: " UINT32_FORMAT_X_0 ")", target_name, (uint)comp_level, hash);
 #ifdef ASSERT
-    } else {
-      uint name_offset = entry->offset() + entry->name_offset();
-      uint name_size   = entry->name_size(); // Includes '/0'
-      const char* name = _cache->cache_buffer() + name_offset;
-      if (strncmp(target_name, name, name_size) != 0) {
-        assert(false, "AOTCodeCache: saved nmethod's name '%s' is different from '%s', hash: " UINT32_FORMAT_X_0, name, target_name, hash);
-      }
+    } else if (method() != entry->method()) {
```
vnkozlov: As you said we can't use pointers (base could be different).
I suggest to keep original code here.

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -926,18 +926,13 @@ AOTCodeEntry* AOTCodeCache::find_code_entry(const methodHandle& method, uint com
   if (is_on() && _cache->cache_buffer() != nullptr) {
     ResourceMark rm;
     const char* target_name = method->name_and_sig_as_C_string();
-    uint hash = java_lang_String::hash_code((const jbyte*)target_name, (int)strlen(target_name));
+    uint hash = (uint)pointer_delta((address)method(), (address)SharedBaseAddress, 1);
```
vnkozlov: Can you add AOTCacheAccess method for calculating delta during product run? It should be together with other AOTCacheAccess  methods we use here.

src/hotspot/share/code/aotCodeCache.cpp
vnkozlov: May be add `assert(entry->id() == hash`

vnkozlov: An other change could be preserving method's name in AOT cache only in debug VM for this compares. For printing we can get name from `entry->method()->name_and_sig_as_C_string()`.

vnkozlov:
And may be few places where we do printing.

vnkozlov:
Some AOT tests crashed on macOS-aarch64 in GHA testing. Please look.

src/hotspot/share/code/aotCodeCache.cpp
ashu-mehra: > As you said we can't use pointers (base could be different).
I suggest to keep original code here.

But we should have already restored entry->method() correctly based on current runtime's base address. So method() should always be same as entry->method().

ashu-mehra: > An other change could be preserving method's name in AOT cache only in debug VM for this compares. For printing we can get name from entry->method()->name_and_sig_as_C_string().

Right, that was in my list as well. I was thinking of not storing the name at all. But I guess we can keep it for debug builds.

src/hotspot/share/code/aotCodeCache.cpp
ashu-mehra: ok, I will add an api in AOTCacheAccess.

src/hotspot/share/code/aotCodeCache.cpp
vnkozlov: > > As you said we can't use pointers (base could be different).
> > I suggest to keep original code here.
> 
> But we should have already restored entry->method() correctly based on current runtime's base address. So method() should always be same as entry->method().

Good point. I agree with this then. Please, print full name too in the assert.

vnkozlov: May be print both names, saved and from `method()->name_and_sig_as_C_string()` only if they are different.

vnkozlov:
This looks good. I will test it.

> Right, that was in my list as well. I was thinking of not storing the name at all. But I guess we can keep it for debug builds.

Lets do it in separate changes, this is not urgent.

vnkozlov:
Crashes in AOT tests in `AOTCodeEntry::method()`:
```
V  [libjvm.so+0x6a2dae]  AOTCodeEntry::method()+0x2e  (aotCacheAccess.hpp:79)
V  [libjvm.so+0x6bb610]  AOTCodeCache::write_nmethod(nmethod*, bool)+0x1190  (aotCodeCache.cpp:1776)
V  [libjvm.so+0x6bb90b]  AOTCodeCache::store_nmethod(nmethod*, AbstractCompiler*, bool)+0x13b  (aotCodeCache.cpp:1613)
V  [libjvm.so+0xa412ea]  ciEnv::register_method(ciMethod*, int, CodeOffsets*, int, CodeBuffer*, int, OopMapSet*, ExceptionHandlerTable*, ImplicitExceptionTable*, AbstractCompiler*, bool, bool, bool, bool, bool, bool, int, bool, AOTCodeEntry*)+0x59a  (ciEnv.cpp:1260)
```

ashu-mehra:
> Lets do it in separate changes, this is not urgent.

Ok

> Crashes in AOT tests in AOTCodeEntry::method()

Yes you are right, the AOTCodeEntry::method() would work only in production run because of SharedBaseAddress. We can fix it in two ways:
1. Remove the usage of method() in assembly phase and add assert in method() that it is being used in production run only. There are currently two usages of method() in assembly run, and both are for null checks.
2. Based on whether it is in assembly run or prod run, use `MetaspaceShared::requested_base_address()` or `SharedBasedAddress`.

I am inclined to 1 because current usage of method() in assembly phase is only for null checks which should always be false.

vnkozlov:
I submitted new testing.

vnkozlov (approved):
My testing passed.

PR #95, opened 2025-09-03
Refactor AOTCodeCache layout to store preload entries separately

ashu-mehra:
Currently the mechanism to lay out final AOTCodeCache entries in `AOTCodeCache::finish_write()` is a bit convoluted.
Code data is initially written in a temporary buffer and then assembled in the final buffer in `AOTCodeCache::finish_write()`.

In the temporary buffer AOTCodeEntry structs are added from the end of the buffer, and the payload (the actual compiled code) is added from the start of the buffer. That means the temporary buffer holds AOTCodeEntry in reverse order.

```
| payload | ... | ACE[n] | ACE[n-1] | ... | ACE[0] |
```

When assembling the final buffer, AOTCodeEntry structs are first copied in the temporary buffer to make the order correct:

```
| payload | ...| ACE[0] | ACE[1] | ... | ACE[n] |... | ACE[n] | ACE[n-1] | ... | ACE[0] |
```

and then the whole memory block is copied into the final buffer.
This means the size of the temporary buffer needs to be a bit more than required.

Another issue is the search table created in `finish_write`. This table includes entries marked for preload. However, preload entries are never searched; they get loaded at the start of the JVM in `preload_aot_code()`. Since the preload and other code entries are mixed together,  we also need a separate table to identify the preload entries.

This PR is an attempt to fix above issues. It does final assembly in following steps:
1. Process AOTCodeEntry structs in the temporary buffer in reverse order and write the ones marked for preload in the final buffer
2. Now the payload for the preload entries is marked
3. Next, add the AOTCodeEntry structs for non-preload code to the final buffer
4. Then add the payload for these entries
5. Finally add the search table

```
| ACE[0] | ... | ACE[m] | payload | ACE[0] | ... | ACE[n] | payload | search_table |
```

This layout separates the preload entries from rest of the code and these entries can then be processed sequentially when the cache is loaded. There is no need for a separate table to identify the preload entries.

I have added the new functionality in separate methods suffixed with `_new` (eg `finish_write_new` and `preload_aot_code_new`) and they are guarded by `UseNewCode` flag.

### DIFF
--- a/src/hotspot/share/code/aotCodeCache.cpp
+++ b/src/hotspot/share/code/aotCodeCache.cpp
@@ -1264,6 +1264,7 @@ bool AOTCodeCache::finish_write() {
     uint new_entries_offset = current - start;
     entries_size = entries_count * sizeof(AOTCodeEntry); // New size
     copy_bytes((_store_buffer + entries_offset), (address)current, entries_size);
+    mark_method_pointer((AOTCodeEntry*)current, entries_count);
     current += entries_size;
 
     log_stats_on_exit();
--- a/src/hotspot/share/code/aotCodeCache.cpp
+++ b/src/hotspot/share/code/aotCodeCache.cpp
@@ -1246,7 +1246,7 @@ bool AOTCodeCache::finish_write() {
         } else if (kind == AOTCodeEntry::Stub) {
           stubs_count++;
         } else {
-          assert(kind == AOTCodeEntry::Code, "sanity");
+          assert(kind == AOTCodeEntry::Nmethod, "sanity");
           nmethods_count++;
         }
       }
--- a/src/hotspot/share/code/aotCodeCache.cpp
+++ b/src/hotspot/share/code/aotCodeCache.cpp
@@ -592,7 +592,11 @@ AOTCodeCache::~AOTCodeCache() {
   // This lock held by ciEnv::register_method() which calls store_nmethod().
   MutexLocker ml(Compile_lock);
   if (for_dump()) { // Finalize cache
-    finish_write();
+    if (UseNewCode) {
+      finish_write_new();
+    } else {
+      finish_write();
+    }
   }
   _load_buffer = nullptr;
   if (_C_store_buffer != nullptr) {
@@ -1066,12 +1070,21 @@ void AOTCodeCache::invalidate_entry(AOTCodeEntry* entry) {
   bool found = false;
   uint i = 0;
   uint count = 0;
-  if (entry->for_preload()) {
-    count = _load_header->preload_entries_count();
-    AOTCodeEntry* preload_entry = (AOTCodeEntry*)addr(_load_header->preload_entries_offset());
-    for (; i < count; i++) {
-      if (entry == &preload_entry[i]) {
-        break;
+  if (UseNewCode) {
+    if (entry->for_preload()) {
+      count = _load_header->preload_entries_count();
+      AOTCodeEntry* preload_entry = (AOTCodeEntry*)addr(_load_header->preload_entries_offset());
+      for (; i < count; i++) {
+        if (entry == &preload_entry[i]) {
+          break;
+        }
+      }
+    } else {
+      count = _load_header->entries_count();
+      for(; i < count; i++) {
+        if (entry == &(_load_entries[i])) {
+          break;
+        }
       }
     }
   } else {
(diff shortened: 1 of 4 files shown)


src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -1266,7 +1291,7 @@ bool AOTCodeCache::finish_write() {
     copy_bytes((_store_buffer + entries_offset), (address)current, entries_size);
     current += entries_size;
 
-    log_stats_on_exit();
+    //log_stats_on_exit();
```
adinn: Not sure if this matters since I assume this method is going to be deleted but why comment this out?

ashu-mehra: you are right, this version of `finish_write()` would go away. I was just lazy to update the code to pass the parameter that `log_stats_on_exit()` needs :)

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -1296,6 +1321,173 @@ bool AOTCodeCache::finish_write() {
...
+      if (size > max_size) {
+        max_size = size;
+      }
+      copy_bytes((_store_buffer + entry->offset()), (address)current, size);
+      entry->set_offset(current - start); // New offset
+      current += size;
+    }
+
+    if (preload_entries_cnt == 0 && entries_count == 0) {
+      log_info(aot, codecache, exit)("AOT Code Cache was not created: no entires");
```
adinn:
```suggestion
      log_info(aot, codecache, exit)("AOT Code Cache was not created: no entries");
```

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -507,12 +508,12 @@ AOTCodeCache::AOTCodeCache(bool is_dumping, bool is_using) :
       return;
     }
     log_info (aot, codecache, init)("Loaded %u AOT code entries from AOT Code Cache", _load_header->entries_count());
-    log_debug(aot, codecache, init)("  Adapters: total=%u", _load_header->adapters_count());
-    log_debug(aot, codecache, init)("  Shared Blobs: total=%u", _load_header->shared_blobs_count());
-    log_debug(aot, codecache, init)("  C1 Blobs: total=%u", _load_header->C1_blobs_count());
-    log_debug(aot, codecache, init)("  C2 Blobs: total=%u", _load_header->C2_blobs_count());
-    log_debug(aot, codecache, init)("  Stubs:    total=%u", _load_header->stubs_count());
-    log_debug(aot, codecache, init)("  Nmethods: total=%u", _load_header->nmethods_count());
+    log_debug(aot, codecache, init)("  %s: total=%u", aot_code_entry_kind_name[AOTCodeEntry::Adapter], _load_header->adapters_count());
```
adinn, line 511: You need to fix the exit trace at line 1300 onwards to call `aot_code_entry_kind_name()`

shipilev:
`UseNewCode` should really be folded. It is good for experimentation, but the final commit should just do the new code unconditionally.

src/hotspot/share/code/aotCodeCache.cpp
ashu-mehra: Those exit traces are in older version of finish_write, which I am going to remove now.

ashu-mehra:
@shipilev I have removed the older version of the functions.

src/hotspot/share/code/aotCodeCache.hpp
```diff
@@ -396,12 +398,14 @@ class AOTCodeCache : public CHeapObj<mtCode> {
...
     uint stubs_count()    const { return _stubs_count; }
-    uint nmethods_count() const { return _entries_count
+    uint nmethods_count() const { uint count = _entries_count
                                        - _stubs_count
                                        - _shared_blobs_count
                                        - _C1_blobs_count
                                        - _C2_blobs_count
-                                       - _adapters_count; }
+                                       - _adapters_count;
+                                  if (UseNewCode) count += _preload_entries_count;
```
vnkozlov: We should not use experimental `UseNewCode` in final PR

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -1167,89 +1173,96 @@ bool AOTCodeCache::finish_write() {
...
-      if (entry->not_entrant()) {
-        log_info(aot, codecache, exit)("Not entrant new entry comp_id: %d, comp_level: %d, hash: " UINT32_FORMAT_X_0 "%s",
-                                       entry->comp_id(), entry->comp_level(), entry->id(), (entry->has_clinit_barriers() ? ", has clinit barriers" : ""));
-        if (entry->for_preload()) {
+      if (entry->for_preload()) {
+        if (entry->not_entrant()) {
           // Skip not entrant preload code:
           // we can't pre-load code which may have failing dependencies.
-          continue;
+          log_info(aot, codecache, exit)("Not entrant new entry comp_id: %d, comp_level: %d, hash: " UINT32_FORMAT_X_0 "%s",
```
vnkozlov: Update message: "Skip not entrant preload code comp_id: ..."

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -1167,89 +1173,96 @@ bool AOTCodeCache::finish_write() {
...
         }
+      }
+    }
+
+    // Now write the data for preload AOTCodeEntry
+    for (int i = 0; i < (int)preload_entries_cnt; i++) {
+      AOTCodeEntry* entry = &preload_entries[i];
+      uint size = align_up(entry->size(), DATA_ALIGNMENT);
+      if (size > max_size) {
+        max_size = size;
```
vnkozlov, line 1211: May be use separate `preload_max_size`

ashu-mehra: `max_size` is only used for logging maximum entry size. I think this is fine, unless we want to report largest preload entry as well.

src/hotspot/share/code/aotCodeCache.cpp
ashu-mehra: Done

src/hotspot/share/code/aotCodeCache.hpp
ashu-mehra: Done

ashu-mehra:
@vnkozlov pushed new commit to address the review comments.

vnkozlov:
My testing passed with known failures and timeouts.
Good.

ashu-mehra:
Thanks @adinn and @vnkozlov for the reviews.

PR #97, opened 2025-09-19
8368095: [leyden] Fix klass dependency recording

shipilev:
See the bug for symptoms. https://github.com/openjdk/leyden/commit/7b7648a4c9f67be509c6fccbcbc0502648388fdc exposed that our dependency recording is not actually accurate. Now that we check that precompiled tasks have dependencies recorded, before we treat class IK as fully initialized, _missing dependencies_ lead to premature replacement of AP4 -> A4, and potential trap from A4.

I see there are several missing things:
 1. We need to record dependencies on object klasses as well, not only on plain metadata.
 2. Since ciObjectFactory is shared, we need to notice objects/metadata even on cache hits.
 3. It looks like we need to observe dependencies even when `cik->is_initialized()` returns `false` at the moment.

Additional testing:
 - [x] Benchmarks

---------
### Progress

### Issue
 * [JDK-8368095](https://bugs.openjdk.org/browse/JDK-8368095): [leyden] Fix klass dependency recording (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/97/head:pull/97` \
`$ git checkout pull/97`

`$ git checkout pull/97` \
`$ git pull https://git.openjdk.org/leyden.git pull/97/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 97`

`$ git pr show -t 97`

</details>

<a href="https://git.openjdk.org/leyden/pull/97.diff">https://git.openjdk.org/leyden/pull/97.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/97#issuecomment-3312234510)
</details>

### DIFF
--- a/src/hotspot/share/code/aotCodeCache.cpp
+++ b/src/hotspot/share/code/aotCodeCache.cpp
@@ -116,6 +116,7 @@ const char* aot_code_entry_kind_name[] = {
 };
 
 static elapsedTimer _t_totalLoad;
+static elapsedTimer _t_totalPreload;
 static elapsedTimer _t_totalRegister;
 static elapsedTimer _t_totalFind;
 static elapsedTimer _t_totalStore;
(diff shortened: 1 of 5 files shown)


shipilev:
Javac benchmark:

```
Benchmark 1: build/linux-x86_64-server-release/images/jdk/bin/java -Xms64m -Xmx1g -XX:+UseSerialGC \
  -cp JavacBenchApp.jar -XX:AOTCache=app.aot JavacBenchApp 50

# 506af854234ed9eeec5543e60f2a0d41b73bc93d
  Time (mean ± σ):     331.7 ms ±   2.3 ms    [User: 660.7 ms, System: 104.6 ms]
  Range (min … max):   328.4 ms … 335.2 ms    10 runs

# current premain
  Time (mean ± σ):     567.5 ms ±  16.5 ms    [User: 1757.1 ms, System: 149.9 ms]
  Range (min … max):   547.0 ms … 599.3 ms    10 runs

# this PR
  Time (mean ± σ):     355.1 ms ±   2.4 ms    [User: 692.8 ms, System: 100.7 ms]
  Range (min … max):   351.6 ms … 358.4 ms    10 runs
```

shipilev:
quarkus-getting-started:

```
$ taskset -c 0-0 make PM_NEW=~/trunks/shipilev-leyden/build/linux-x86_64-server-release/images/jdk/ PM_OLD=~/trunks/shipilev-leyden/build/baseline/ BLDJDK_HOME=~/Install/jdk21u-ea/ MAINLINE_JAVA=~/Install/jdk21u-ea/bin/java compare_premain_builds

Run,Old CDS + AOT,New CDS + AOT
1,425,318
2,415,315
3,410,319
4,421,310
5,424,311
6,412,332
7,411,326
8,422,319
9,413,334
10,431,320
Geomean,418.35,320.31 (1.31x improvement)
Stdev,6.79,7.66
```

iwanowww:
> It looks like we need to observe dependencies even when cik->is_initialized() returns false at the moment.

Can you elaborate more on it? Does it relate to the case when corresponding class is being initialized?

shipilev:
> Just out of interest, what's the difference if you add those object/klass deps only when they are new rather than already in the cache?

Ad-hoc culling of this PR shows this:

```
Uninitialized /  total traps:

 553 / 691  Current premain
 138 / 276  This PR, but gating IK->is_initializing for deps recording [see above comment for fuller result]
  45 / 179  This PR, but skipping recording on cache hits
   2 / 129  This PR, but skipping object recording
   2 / 118  This PR, full version
```

The performance improves as we go from top to bottom. Arguably, this might suggest we do not need to handle oop accesses, really, but there is a minor decrease to the number of total traps and what _looks to be_ a minor performance improvement on top.

iwanowww:
Thanks for the clarifications, Aleksey.

There's a long-standing issue with how nmethod init dependencies are managed: they are recorded during training run and then left intact throughout assembly phase while compilers try hard to adapt to the exact set of init dependences during archived code generation. 

Ideally, JITs should produce a new set of init dependences based on training data and observations made during AOT compilation. When it comes to init dependencies, there's an important distinction between classes which were eventually initialized and those which were never initialized during training run. In the former case, it's better to record an extra dependency and delay A4 nmethod installation while in the former case an uncommon trap is preferred. (Similar reasoning applies to classes in being_initialized state: if class initialization never finishes, there's no point in recording an init dependency for it.)

For now, proposed fix looks good.

iwanowww (approved):

PR #99, opened 2025-09-23
8368465: [leyden] Improve precompiler method selection code

shipilev:
Forked from [JDK-8366681](https://bugs.openjdk.org/browse/JDK-8366681): there are still some cleanups/performance improvements possible. Current selection code is a bit hairy, and turns out the changes I made for previous patch improve performance.

Notable improvements:
 1. Push the compilation level filters downwards. This allows compiling A2 from T2/T3 code more easily, and allows to implement policies for compiling on any A* level based on observing top-compiled T* levels.
 2. Sort methods by hotness and code size. This looks to have a positive effect on shorter workloads, I suspect because we are avoiding a lot of C1 compilations by preloading hottest code first.

Additional testing:
 - [x] Performance tests (see comments)

---------
### Progress

### Issue
 * [JDK-8368465](https://bugs.openjdk.org/browse/JDK-8368465): [leyden] Improve precompiler method selection code (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/99/head:pull/99` \
`$ git checkout pull/99`

`$ git checkout pull/99` \
`$ git pull https://git.openjdk.org/leyden.git pull/99/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 99`

`$ git pr show -t 99`

</details>

<a href="https://git.openjdk.org/leyden/pull/99.diff">https://git.openjdk.org/leyden/pull/99.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/99#issuecomment-3323846322)
</details>

### DIFF
--- a/src/hotspot/share/compiler/precompiler.cpp
+++ b/src/hotspot/share/compiler/precompiler.cpp
@@ -33,7 +33,6 @@
 #include "compiler/precompiler.hpp"
 #include "logging/logStream.hpp"
 #include "memory/allocation.hpp"
-#include "oops/methodCounters.hpp"
 #include "oops/trainingData.hpp"
 #include "runtime/handles.inline.hpp"
 
@@ -94,8 +93,8 @@ public:
   void operator()(TrainingData* td) {
     if (td->is_MethodTrainingData()) {
       MethodTrainingData* mtd = td->as_MethodTrainingData();
-      if (mtd->has_holder() && include((Method*)mtd->holder())) {
-        _methods.push((Method*)mtd->holder());
+      if (mtd->has_holder() && include(mtd->holder())) {
+        _methods.push(mtd->holder());
       }
     }
   }


shipilev:
Ready for review, folks. There are performance benefits of doing this, very apparently.

src/hotspot/share/compiler/precompiler.cpp
```diff
@@ -56,12 +56,30 @@ class PrecompileIterator : StackObj {
...
+      case CompLevel_full_optimization:
+        // For final C1/C2 compilations, we only compile when there was relevant compilation during training.
+        return _comp_level == high_top_level;
+      case CompLevel_limited_profile:
+        // For profiled C1 compilations, generate limited profile when there was limited/full
+        // profiled compilation in training.
+        return CompLevel_limited_profile <= high_top_level && high_top_level <= CompLevel_full_profile;
+      case CompLevel_full_profile:
+        // We do not include C1 full profiled methods at this time.
+        // TODO: See if it is profitable to do so. This requires MDO support in AOTCache.
```
vnkozlov: We already support reference to MDO from C1 compiled code because we cache tier2 which do profiling.
But we may miss few places in tier3 code.

src/hotspot/share/compiler/precompiler.cpp
```diff
@@ -82,48 +100,52 @@ class PrecompileIterator : StackObj {
...
 
-  static int compare_by_compile_id_tier4(Method** m1, Method** m2) {
-    return compare_by_compile_id(m1, m2, CompLevel_full_optimization);
+  static size_t counts(Method* m) {
+    size_t count = 0;
+    MethodTrainingData* mtd = method_training_data(m);
+    if (mtd != nullptr) {
+      MethodData* md = mtd->final_profile();
+      if (md != nullptr) {
+        count += md->backedge_count();
```
vnkozlov: Hmm, this will put methods with hot loop up front.

src/hotspot/share/compiler/precompiler.cpp
```diff
@@ -82,48 +100,52 @@ class PrecompileIterator : StackObj {
...
+  static int compare_methods(Method** m1, Method** m2) {
+    // Hottest methods go first.
+    size_t c1 = counts(*m1);
+    size_t c2 = counts(*m2);
+    if (c1 > c2) return -1;
+    if (c1 < c2) return +1;
 
-      default: fatal("%d", _search_level);
-    }
+    // Otherwise, break the tie by code size: largest methods go first.
```
vnkozlov: What is the reason for larger method be first? Can we use compile ID here instead?

src/hotspot/share/compiler/precompiler.cpp
shipilev: I thought we only record MCS, and reinstantiating MDO. But my recollection is vague. I dropped this to avoid confusion.

src/hotspot/share/compiler/precompiler.cpp
shipilev: Yes, this is intentional: this effectively puts the methods that are profitable to (pre)load first, so they: a) do not linger in interpreter too much; b) do not trigger JIT compilation before AOT code is able to (pre)load. The methods with hot back-branches are those methods :)

src/hotspot/share/compiler/precompiler.cpp
shipilev: So my logic was like with the hot methods. If we have lost the game of "preload the AOT code before JIT got triggered", and JIT got triggered, we want to then prioritize larger methods, as they are more likely to take more time to JIT compile. In other words, I think if you lost to JIT timing-wise, you want to preempt the fattest JIT compiles first. But it is only a bet. If we ever record compilation time in nmethods/profiles, we could have used that to break the tie.

vnkozlov: I am not sure understand how different order of pre/AOT-compilation can affect performance of production run.
We bulk load all "Preload" AOT code - ordering does not matter for it. Even if we load in selected order. It is one thread which do loading and it is blocking (I actually playing with spreading this preload on all compiler threads - did not see much affect on startup).

The only explanation is that preload happens only when C2 compiler threads are initialized (Preload AOT code is C2 compiled code) and it happens simultaneously with C1 threads initialization which could be available sooner for C1 compilation than we finish preloading. Especially on small number of cores machines.  I did observed that we start loading A1 and A2 code first (even normal C1 compilations) before we start preload AP4. 

Is it what you are trying to solve here?

The invocation counters should be roughly the same for methods without loops (10000 to trigger C2 compilation). They could be different if code was deoptimized and run in interpreter.  The only difference id backedge counter. So in this sense you push methods with hot loop up front as we talked about in other comment. Which may affect performance but it would depend on application.

I agree with ordering by size (or time spant in compilation) but only for methods which did not have A1 or A2 code. Which should not be the case - if we have AP4 we will have A1 and A2 for it. I am still not convince by this.

May be  we should try to move `AOTCodeCache::preload_code()` just after `SystemDictionary::compute_java_loaders()` because it does not depend on training data. So we can have AP4 sooner.  MMaping directly into CodeCache will also speedup preloading - it is on our list to do.

vnkozlov:
An other suggestion for this concurrent preloading would be to split A4 preload code. One set is the current which needs to wait `compute_java_loaders()`. And new one (much smaller) is for simple methods for classes which are loaded first (String, for example) which we can preload much sooner.

src/hotspot/share/compiler/precompiler.cpp
shipilev: > We bulk load all "Preload" AOT code - ordering does not matter for it. Even if we load in selected order. It is one thread which do loading and it is blocking (I actually playing with spreading this preload on all compiler threads - did not see much affect on startup).

Um, I don't think that's the case? Preloading is asynchronous. See `CompileBroker::compile_method`:

```
    bool is_blocking = ReplayCompiles                                             ||
                       !directive->BackgroundCompilationOption                    ||
                       (PreloadBlocking && (compile_reason == CompileTask::Reason_Preload));
    compile_method_base(method, osr_bci, comp_level, hot_count, compile_reason, requires_online_compilation, is_blocking, THREAD);
```

We have the option to _make_ preload blocking (`PreloadBlocking`), but it is turned off by default. We know enabling `+PreloadBlocking` is counter-productive, because it could easily take hundreds of milliseconds. 

So while compilers are working through preloading the code, the application runs and can trigger compilations. Sorting preloading methods allows loading hottest code before that code transits to normal compilation. This is the problem I am trying to mitigate.

Maybe the compilation policy should actually participate in preloading: i.e. if there is a hot code that transitions from T0 to any other level, attempt the preload first, in case normal preloading is lagging behind. That would be more intrusive, though, so as the conservative approach I would like to prioritize more profitable preload code first.

vnkozlov: > So while compilers are working through preloading the code, the application runs and can trigger compilations. Sorting preloading methods allows loading hottest code before that code transits to normal compilation. This is the problem I am trying to mitigate.

Okay, I think I can understand when compile ID may screw up us. If T4 compilation during training run happens several times for the same method due to deoptimization we will cache only last corresponding AP4:
```
      if (entry->for_preload()) {
        if (entry->not_entrant()) {
          // Skip not entrant preload code:
```
such entry will have high compile ID. We can keep early ID and used it for cached AP4 to avoid this.

Which leads to an other issue. In initial AOT code implementation I kept deoptimization counter for A4 and use it when search A4 to load in production run. We removed that counter but kept all versions of A4 but `find_entry()` will return first A4 it found which may have a lot more uncommon traps when latest A4 version. May be we should filter A4 the same way we do for AP4.

vnkozlov: By "blocking" I mean that we have only one AOT compiler thread to load AP4.

vnkozlov: May be your change reduced number of AOT compiled nmethod in cache which allow faster processing.
Please run with `-Xlog:aot+codecache+init=debug -XX:+CITime` for production run to see how many AOT nmethods in AOT cache and how many were loaded/used.

shipilev: Actually... Now I see we generate, and thus use substantially A2 code! This also aligns with performance data: we have way fewer C1 compilations with this patch.

```
# ==== Baseline

Create:
  [4.593s][info][precompile] Precompilation for level 1 finished (94 successful out of 94 total)
  [4.604s][info][precompile] Precompilation for level 2 finished (131 successful out of 131 total)
  [4.814s][info][precompile] Precompilation for level 2 finished (1852 successful out of 1852 total)
  [6.035s][info][precompile] Precompilation for level 4 finished (1660 successful out of 1660 total)
  [4.589s][info][precompile] Precompilation for level 5 finished (1660 successful out of 1660 total)

Use:
  Tier1 {speed: 42838.159 bytes/s; standard:  0.014 s, 582 bytes, 135 methods; ...}
  Tier2 {speed: 210303.802 bytes/s; standard:  0.311 s, 63857 bytes, 817 methods; ...}
  Tier3 {speed: 134013.414 bytes/s; standard:  0.035 s, 4685 bytes, 245 methods; ...}
  Tier4 {speed: 69205.374 bytes/s; standard:  0.051 s, 3225 bytes, 13 methods; ...}
  AOT Code T1 {speed: 297580.645 bytes/s; standard:  0.001 s, 369 bytes, 94 methods; ...}
  AOT Code T2 {speed: 5654043.587 bytes/s; standard:  0.042 s, 237861 bytes, 1969 methods; ...}
  AOT Code T4 {speed: 25219362.296 bytes/s; standard:  0.029 s, 737408 bytes, 927 methods; ...}
  AOT Code T5 {speed: 30793594.418 bytes/s; standard:  0.048 s, 1474270 bytes, 1658 methods; ...}


# ==== Patched

Create:
  [3.984s][info][precompile] Precompilation for level 1 finished (311 successful out of 311 total)
  [4.382s][info][precompile] Precompilation for level 2 finished (2752 successful out of 2752 total)
  [4.383s][info][precompile] Precompilation for level 3 finished (0 successful out of 0 total)
  [5.392s][info][precompile] Precompilation for level 4 finished (1641 successful out of 1641 total)
  [3.972s][info][precompile] Precompilation for level 5 finished (1641 successful out of 1641 total)

Use:
  Tier1 {speed:  0.000 bytes/s; standard:  0.000 s, 0 bytes, 0 methods; ...
  Tier2 {speed: 579987.470 bytes/s; standard:  0.026 s, 15526 bytes, 44 methods; ...
  Tier3 {speed: 181499.273 bytes/s; standard:  0.026 s, 4761 bytes, 254 methods; ...
  Tier4 {speed: 77265.133 bytes/s; standard:  0.027 s, 2087 bytes, 12 methods; ...
  AOT Code T1 {speed: 432360.583 bytes/s; standard:  0.002 s, 942 bytes, 228 methods; ...
  AOT Code T2 {speed: 6664604.248 bytes/s; standard:  0.042 s, 281287 bytes, 2735 methods; ...
  AOT Code T4 {speed: 26296881.658 bytes/s; standard:  0.026 s, 682331 bytes, 924 methods; ...
  AOT Code T5 {speed: 33814172.284 bytes/s; standard:  0.042 s, 1430045 bytes, 1632 methods; ...
```

So I have changed something in selection code that takes on more A2 compiles, profitably.  Have not yet confirmed if preload order has any effect on top of that. Investigating...

shipilev: LOL, I think I found the performance bug in the original code that I fixed by accident. `MTD::highest_level()` includes the cases when method is inlined. For T2/T3 code, it would return `4` if we ended up inlining that method into T4 code. Which would fail the inclusion check for A2 compilation, even though we did have a legit top-level T2/T3 compile. I would say the fact we have inlined T2/T3 in _some_ T4 should _not_ disqualify A2 compilation. 

This change alone gives the same kind of performance boost as my patch:

```
diff --git a/src/hotspot/share/compiler/precompiler.cpp b/src/hotspot/share/compiler/precompiler.cpp
index 04f95857a63..8a5da803b04 100644
--- a/src/hotspot/share/compiler/precompiler.cpp
+++ b/src/hotspot/share/compiler/precompiler.cpp
@@ -84,7 +84,7 @@ class PrecompileIterator : StackObj {
 
   static int compile_id(Method* m, int level) {
     MethodTrainingData* mtd = m->method_holder()->is_loaded() ? MethodTrainingData::find(methodHandle(Thread::current(), m)) : nullptr;
-    if (mtd != nullptr && mtd->highest_level() == level) {
+    if (mtd != nullptr && mtd->highest_top_level() == level) {
       CompileTrainingData* ctd = mtd->last_toplevel_compile(level);
       if (ctd != nullptr) {
         return ctd->compile_id();
```

This thing is confusing, and yet another reason why I this PR looks more understandable: it very explicitly checks `MTD::highest_top_level()` when deciding whether to accept the method. It does not do this with this bug, and also does not implicitly participate in filtering by `compile_id() < INT_MAX`.

shipilev: Reverted the sorting back to compile IDs instead of counters/size, as it does not affect performance, really.

shipilev:
Any news on testing, @vnkozlov?

PR #100, opened 2025-09-24
8368595: [leyden] Enhance CPU features check

Hamlin-Li:
Hi,
Can you help to review this patch?

Current implementation of CPU features check compare the whole buffer of cpu features. This could introduce a potential issue when CPU features accumulate.

Consider this example.
If at JDK version X there are 5 CPU features, they are a(0), b(1), c(2), d(3), e(4), the number inside () is the feature index assigned (it could be explicitly or implicitly, depends on the implementation of VM_Version). In the future at JDK version Y there could be another feature f added, but f could be added at the middle of features list (reason could be to group some features together or order all cpu features in alphabetical way, so now it looks like a(0), b(1), c(2), f(3), d(4), e(5).
So, cpu features buffer of the version X in an archive could be 111110000..., which means all 5 features (a/b/c/d/e) are supported, and this archive could be used by a runtime of version Y, and the local CPU supports a/b/c/f/d, but not e, in this situation, the cpu features are also 111110000..., which means buffer comparison result is equal, but indeed they are not equal.

The solution could be to just introduce an cpu_feature_number, before comparing the buffer, compare if the number is equal, if not just fail the cpu feature comparison.
This solution is based on the assumption that we only add but not remove CPU features.

Thanks!

---------
### Progress

### Issue
 * [JDK-8368595](https://bugs.openjdk.org/browse/JDK-8368595): [leyden] Enhance CPU features check (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/100/head:pull/100` \
`$ git checkout pull/100`

`$ git checkout pull/100` \
`$ git pull https://git.openjdk.org/leyden.git pull/100/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 100`

`$ git pr show -t 100`

</details>

<a href="https://git.openjdk.org/leyden/pull/100.diff">https://git.openjdk.org/leyden/pull/100.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/100#issuecomment-3330508207)
</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
@@ -761,7 +761,7 @@ void VM_Version::store_cpu_features(void* buf) {
   *(uint64_t*)buf = _features;
 }
 
-bool VM_Version::supports_features(void* features_buffer) {
+bool VM_Version::supports_features(void* features_buffer, uint cpu_features_number) {
   uint64_t features_to_test = *(uint64_t*)features_buffer;
   return (_features & features_to_test) == features_to_test;
 }
--- a/src/hotspot/cpu/aarch64/vm_version_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/vm_version_aarch64.hpp
@@ -235,7 +235,7 @@ enum Ampere_CPU_Model {
   // Size of the buffer must be same as returned by cpu_features_size()
   static void store_cpu_features(void* buf);
 
-  static bool supports_features(void* features_to_test);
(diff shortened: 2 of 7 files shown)


ashu-mehra:
@Hamlin-Li while I don't disagree with adding a check for cpu features number as such, I think the scenario you mentioned is not a problem because AOTCache is tightly coupled to the JDK version. So an AOTCache built with a JDK version X that supports 5 features cannot be used with a JDK version Y with support for 6 features.

PR #101, opened 2025-09-26
Minor cleanup and refactoring

ashu-mehra:
Removed unused parameter `aot_code_entry` in `ciEnv::register_method`.
Removed `preload` local variable in `ciEnv::register_method` as it is always false.
Refactored `ciEnv::is_compilation_valid` a bit.

### DIFF
--- a/src/hotspot/share/ci/ciEnv.cpp
+++ b/src/hotspot/share/ci/ciEnv.cpp
@@ -983,26 +983,14 @@ void ciEnv::validate_compile_task_dependencies(ciMethod* target) {
   }
 }
 
-// aot_code_entry != nullptr implies loading compiled code from AOT code cache
-bool ciEnv::is_compilation_valid(JavaThread* thread, ciMethod* target, bool preload, bool install_code, CodeBuffer* code_buffer, AOTCodeEntry* aot_code_entry) {
+// is_aot_code = true implies loading compiled code from AOT code cache
+bool ciEnv::is_compilation_valid(JavaThread* thread, ciMethod* target, bool install_code, bool is_aot_code, bool preload) {
   methodHandle method(thread, target->get_Method());
 
   // We require method counters to store some method state (max compilation levels) required by the compilation policy.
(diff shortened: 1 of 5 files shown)


src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -983,28 +983,16 @@ void ciEnv::validate_compile_task_dependencies(ciMethod* target) {
...
-      code_buffer->free_blob();
-    }
     return false;
   }
 
-  if (aot_code_entry != nullptr) {
-    // Invalid compilation states:
-    //  - AOTCodeCache is closed, AOTCode entry is garbage.
-    //  - AOTCode entry indicates this shared code was marked invalid while it was loaded.
-    if (!AOTCodeCache::is_on() || aot_code_entry->not_entrant()) {
```
ashu-mehra, line 1003: We don't need to check for `AOTCodeCache::is_on` anymore because AOTCodeCache won't be closed if there are nmethod readers. See `AOTCodeCache::wait_for_no_nmethod_readers`.

src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -983,28 +983,16 @@ void ciEnv::validate_compile_task_dependencies(ciMethod* target) {
   }
 }
 
-// aot_code_entry != nullptr implies loading compiled code from AOT code cache
-bool ciEnv::is_compilation_valid(JavaThread* thread, ciMethod* target, bool preload, bool install_code, CodeBuffer* code_buffer, AOTCodeEntry* aot_code_entry) {
+// is_aot_code = true implies loading compiled code from AOT code cache
```
vnkozlov: Suggestion: `is_aot_code` -> `is_loading_aot_code`

vnkozlov (commented):
Just one comment

src/hotspot/share/ci/ciEnv.cpp
ashu-mehra: Done

vnkozlov (approved):
Good

PR #102, opened 2025-09-27
8368811: [Leyden] Use AOTRuntimeConstants table for card_table::_byte_map_base

vnkozlov:
Currently we use `Relocation` info to patch `card_table::_byte_map_base` referenced in compiled code. This is not completely correct since this is not address. We currently have special check in `AOTCodeCache::init2()` to skip AOT code generation and usage if `byte_map_base` is not relocatable [aotCodeCache.cpp#L341](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/code/aotCodeCache.cpp#L341)

To avoid this we should use existing `AOTRuntimeConstants` table to load `byte_map_base` from it.

I also added few missing loads `card_shift` from `AOTRuntimeConstants` table. Actually I am not sure why we have it in `AOTRuntimeConstants` table.  From what I see, it is based on `GCCardSizeInBytes` flag [cardTable.cpp#L46](https://github.com/openjdk/leyden/blob/premain/src/hotspot/share/gc/shared/cardTable.cpp#L46) and never modified. The flags is not adjusted ergonomically.  So we can simple record the flag in AOT code configuration and verify it when loading AOT code. 

@adinn  what do you think about `card_shift` in `AOTRuntimeConstants` table? You added it there.

Changes were testing tier1-5.

### DIFF
--- a/src/hotspot/cpu/aarch64/gc/shared/cardTableBarrierSetAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/gc/shared/cardTableBarrierSetAssembler_aarch64.cpp
@@ -23,6 +23,7 @@
  */
 
 #include "asm/macroAssembler.inline.hpp"
+#include "code/aotCodeCache.hpp"
 #include "gc/shared/barrierSet.hpp"
 #include "gc/shared/cardTable.hpp"
 #include "gc/shared/cardTableBarrierSet.hpp"
@@ -37,8 +38,20 @@ void CardTableBarrierSetAssembler::store_check(MacroAssembler* masm, Register ob
   BarrierSet* bs = BarrierSet::barrier_set();
   assert(bs->kind() == BarrierSet::CardTableBarrierSet, "Wrong barrier set kind");
 
-  __ lsr(obj, obj, CardTable::card_shift());
-
+#if INCLUDE_CDS
+  // AOT code needs to load the barrier card shift from the aot
+  // runtime constants area in the code cache otherwise we can compile
+  // it as an immediate operand
+  if (AOTCodeCache::is_on_for_dump()) {
+    address card_shift_address = (address)AOTRuntimeConstants::card_shift_address();
+    __ lea(rscratch1, ExternalAddress(card_shift_address));
+    __ ldrb(rscratch1, rscratch1);
+    __ lsrv(obj, obj, rscratch1);
+  } else
+#endif
+  {
+    __ lsr(obj, obj, CardTable::card_shift());
+  }
   assert(CardTable::dirty_card_val() == 0, "must be");
 
   __ load_byte_map_base(rscratch1);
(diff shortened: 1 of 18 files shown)


vnkozlov:
I updated changes.

src/hotspot/cpu/aarch64/gc/shared/cardTableBarrierSetAssembler_aarch64.cpp
```diff
@@ -32,25 +32,25 @@
...
 
   assert(CardTable::dirty_card_val() == 0, "must be");
 
-  __ load_byte_map_base(rscratch1);
+  __ load_byte_map_base(rscratch);
 
   if (UseCondCardMark) {
     Label L_already_dirty;
-    __ ldrb(rscratch2,  Address(obj, rscratch1));
+    __ ldrb(rscratch2,  Address(obj, rscratch));
```
adinn, line 49: Do we need an extra precondition?
`precond(!UseCondCardMark || rscratch != rscratch2)`

src/hotspot/share/gc/shared/c1/cardTableBarrierSetC1.cpp
```diff
@@ -76,6 +78,18 @@ void CardTableBarrierSetC1::post_barrier(LIRAccess& access, LIR_Opr addr, LIR_Op
...
 
   LIR_Address* card_addr;
+#if INCLUDE_CDS
+  if (AOTCodeCache::is_on_for_dump()) {
+    // load the card table address from the AOT Runtime Constants area
+    LIR_Opr byte_map_base_adr = LIR_OprFact::intptrConst(AOTRuntimeConstants::card_table_address());
+    LIR_Opr byte_map_base_reg = gen->new_pointer_register();
+    __ move(byte_map_base_adr, byte_map_base_reg);
+    LIR_Address* byte_map_base_indirect = new LIR_Address(byte_map_base_reg, 0, T_LONG);
+    //LIR_Opr byte_map_base = gen->new_pointer_register();
```
adinn: Do we need this comment?

src/hotspot/share/gc/shared/c1/cardTableBarrierSetC1.cpp
```diff
@@ -64,6 +65,7 @@ void CardTableBarrierSetC1::post_barrier(LIRAccess& access, LIR_Opr addr, LIR_Op
   assert(addr->is_register(), "must be a register at this point");
 
 #ifdef CARDTABLEBARRIERSET_POST_BARRIER_HELPER
+  assert(!aotCodeCache::is_on(), "this path is not implemented");
```
adinn:
```suggestion
  assert(!AOTCodeCache::is_on(), "this path is not implemented");
```

vnkozlov: Goot catch. 32-bit ARM build failed because of this.

src/hotspot/share/gc/shared/c1/cardTableBarrierSetC1.cpp
vnkozlov: Removed. I experimented with separate registers but reusing register (byte_map_base_reg) seems work too.

src/hotspot/cpu/aarch64/gc/shared/cardTableBarrierSetAssembler_aarch64.cpp
vnkozlov: added

vnkozlov:
I looked on Shenandoah barriers cpu specific code and don't see any issues. As you said, it loads card table base via a thread local and use constant for card_shift (so added flag check will work for it too).

vnkozlov:
Thank you @adinn 
I will wait for Ioi's mainline merge to premain to integrate this.

PR #117, opened 2026-06-30
8387494: ResolvedMethodName::vmtarget is not updated when class in AOT cache is redefined

iklam:
If we have cached ResolvedMethodName objects, they must be added to the ResolvedMethodTable during bootstrap. If not, the new test case `RedefineAllTest` will fail with an assert.

---------

---------
### Progress

### Issue
 * [JDK-8387494](https://bugs.openjdk.org/browse/JDK-8387494): ResolvedMethodName::vmtarget is not updated when class in AOT cache is redefined (**Bug** - P3)

### Reviewers
 * [Andrew Dinn](https://openjdk.org/census#adinn) (@adinn - **Reviewer**) ⚠️ Review applies to [4cf47ba0](https://git.openjdk.org/leyden/pull/117/files/4cf47ba076f8a16a9727a0316179d290f786b51f)
 * [Igor Veresov](https://openjdk.org/census#iveresov) (@veresov - **Reviewer**) ⚠️ Review applies to [4cf47ba0](https://git.openjdk.org/leyden/pull/117/files/4cf47ba076f8a16a9727a0316179d290f786b51f)
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/117/head:pull/117` \
`$ git checkout pull/117`

`$ git checkout pull/117` \
`$ git pull https://git.openjdk.org/leyden.git pull/117/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 117`

`$ git pr show -t 117`

</details>

<a href="https://git.openjdk.org/leyden/pull/117.diff">https://git.openjdk.org/leyden/pull/117.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/117#issuecomment-4921525204)
</details>


src/hotspot/share/cds/heapShared.cpp
```diff
@@ -1310,9 +1319,27 @@ void HeapShared::serialize_tables(SerializeClosure* soc) {
...
+  if (_runtime_resolved_methods != nullptr) {
+    JavaThread* current = JavaThread::current();
+    HandleMark hm(current);
+    for (int i = 0; i < _runtime_resolved_methods->length(); i++) {
+      int root_index = _runtime_resolved_methods->at(i);
+      Handle mem_name(current,  get_root(root_index, /*clear=*/true));
+      Method* method = java_lang_invoke_ResolvedMethodName::vmtarget(mem_name());
+      InstanceKlass* holder = method->method_holder();
+      holder->set_has_resolved_methods();
+      oop o = ResolvedMethodTable::add_method(method, mem_name);
```
adinn, line 1337: I can understand that this registration of resolved methods is needed to make the agent transform work correctly -- `VM_RedefineClasses::doit` calls `ResolvedMethodTable::adjust_method_entries` when any of the transformed classes has associated resolved methods. However, why was registration not needed for other clients e.g. `ServiceThread`?

src/hotspot/share/oops/trainingData.cpp
```diff
@@ -119,7 +119,7 @@ void TrainingData::verify() {
 }
 
 static bool is_excluded(InstanceKlass* k) {
-  if (!k->is_loaded() || k->has_been_redefined()) {
+  if (!k->is_loaded() || (TrainingData::need_data() && k->has_been_redefined())) {
```
adinn, line 122: n.b. This is more of a side-comment than a review comment.

This fixes the issue I tripped up over a couple of weeks ago when using an agent in a PRODUCTION run. A shutdown hook was performing verification of all training data at the end of the run and was unhappy when the owning class of some MethodTrainignData was excluded because the agent had transformed it. This fix avoids the verify error.

However, I wonder if the shutdown hook really needs to be registered in PRODUCTION? Perhaps when we are doing linear training? But otherwise? @veresov what do you think?

adinn (approved):
Changes look good and tests look fine.

src/hotspot/share/cds/heapShared.cpp
coleenp: The service thread just cleans out stale entries.

veresov (approved):
Looks good!

test/hotspot/jtreg/runtime/cds/appcds/agent/RedefineAllAgent.java
```diff
@@ -0,0 +1,68 @@
+/*
```
coleenp, line 1: There's a RedefineHelper that the serviceability tests use in test/hotspot/jtreg/serviceability/jjvmti/RedefineClasses that does all this setup for you.  All you do is give it a new class in a string.  Do you need all of this?
Seems like you can do AOT save some MemberName, and use the helper to redefine it the class that has the member name method.  You don't even have to change it for the Method to become old.

coleenp (commented):
The change makes sense but the test seems like extra work.

test/hotspot/jtreg/runtime/cds/appcds/agent/RedefineAllAgent.java
iklam: I rewrote two of the tests to use `RedefineClassHelper`. However, for RedefineAllAgent.java, I have to made a call into `Instrumentation.retransformClasses()` because the bytecodes of some of the classes in the AOT cache cannot be found in JAVA_HOME/lib/modules. It's easier to write a stand-lone agent in the case.

test/lib/RedefineClassHelper.java
```diff
@@ -103,6 +110,33 @@ public static byte[] replaceClassName(ClassLoader loader, String oldClassName, S
...
+     * Example:
+     *
+     * RedefineClassHelper.redefineMethodBodies(ShouldBeTransformed.class,
+     *                                          (MethodModel method) -> method.methodName().equalsString("toString"),
+     *                                          (CodeBuilder builder, CodeElement element) -> {
+     *                                              builder.ldc("YYYY");
+     *                                              builder.areturn();
+     *                                          });
+     */
+    public static void redefineMethodBodies(Class clazz, Predicate<MethodModel> filter, CodeTransform xform) throws Exception {
```
coleenp: javac is warning for raw Class without Class<?>.

iklam: I've changed the two uses of "raw' Class to Class<?>.

coleenp (approved):
This looks good.  The RedefineClassHelper.redefineMethodBodies seems like it could be generally helpful, rather than turning the class definition into a string.

iklam:
Thanks @adinn @veresov @coleenp for the review
/integrate

PR #115, opened 2026-06-11
8386588: [leyden] Improve map file logging to add code cache

Delawen:
Mimicking https://github.com/vnkozlov/jdk/pull/1 for the leyden premain2 repository.

Adding to the assembly mapfile the assets of the code cache.

Tests in https://github.com/Delawen/jdk/actions/runs/27264662339

The map file is generated for example with `-Xlog:aot+map=trace,aot+map+oops=trace,aot=warning:file=${FOLDER}/aot.map:none:filesize=0`

Example of output:

For NMethods: 
`Pointer` : @@ `Nmethod` `Size` `Compilation Level` `Compilation ID` `Method signature`

Fort the rest:
`Pointer` : @@ `Kind` `Size` `ID` `Name`

```
0x00007fbcf7ffeff4: @@ StubGenBlob       25896 73 initial_blob (stub gen)
0x00007fbcf7ffefc8: @@ SharedBlob        403 13 throw_StackOverflowError_blob (shared runtime)
0x00007fbcf7ffef9c: @@ StubGenBlob       3621 74 continuation_blob (stub gen)
0x00007fbcf7ffed34: @@ SharedBlob        2156 0 deopt_blob (shared runtime)
0x00007fbcf7ffe49c: @@ StubGenBlob       129373 75 compiler_blob (stub gen)
0x00007fbcf7ffe470: @@ StubGenBlob       59182 76 final_blob (stub gen)
0x00007fbcf7ffcc8c: @@ C1Blob            650 17 dtrace_object_alloc_blob (C1 runtime)
0x00007fbcf7ffcc60: @@ C1Blob            387 18 unwind_exception_blob (C1 runtime)
0x00007fbcf7ffbe74: @@ C2Blob            335 70 vthread_start_transition_blob (C2 runtime)
0x00007fbcf7ffbe48: @@ Adapter           1333 217 ILIL
0x00007fbcf7ffbe1c: @@ Adapter           1376 218 LIIILLL
0x00007fbcf7ffbdf0: @@ C2Blob            333 71 vthread_end_transition_blob (C2 runtime)
0x00007fbcf7ffbdc4: @@ Adapter           1460 219 LLLIILILLII
0x00007fbcf7ffbd98: @@ Adapter           1342 220 LIIIL
[...]

PR #118, opened 2026-07-16
8388410: [Leyden] Improve map file logging to add Embedded Stubs from StubGenBlobs

Delawen:
Continuation of https://github.com/openjdk/leyden/pull/115

Right now, we log StubGenBlobs in premain2: https://github.com/openjdk/leyden/pull/115

Current output example:
```
0x00007f1b37ffefe4: @@ StubGenBlob 24452 73 initial_blob (stub gen)
```

But we are skipping logging of the embedded stubs inside those blobs. This PR improves the logging to add the embedded stubs too.

The log line for `EmbeddedStub` contains the following data:
 - Pointer
 - Size of the Embedded Stub
 - Id of the StubGenBlob (BlobId) to link back to which StubGenBlob it belongs (although it should be ordered)
 - Id of the Embedded Stub
 - Name of the Embedded stub
 
Example of output:
```
0x00007fd2fbffefe4: @@ StubGenBlob       24452 73 initial_blob (stub gen)
0x00007fd2dbfff05b: @@ EmbeddedStub      276 73 77 call_stub_stub (stub gen)
0x00007fd2dbfff010: @@ EmbeddedStub      75 73 78 forward_exception_stub (stub gen)
0x00007fd2dbfff16f: @@ EmbeddedStub      31 73 79 catch_exception_stub (stub gen)
0x00007fd2dbfff350: @@ EmbeddedStub      612 73 80 updateBytesCRC32_stub (stub gen)
0x00007fd2dbfff5d0: @@ EmbeddedStub      612 73 81 updateBytesCRC32C_stub (stub gen)
0x00007fd2dbfff83e: @@ EmbeddedStub      14 73 82 f2hf_stub (stub gen)
0x00007fd2dbfff834: @@ EmbeddedStub      10 73 83 hf2f_stub (stub gen)
0x00007fd2dc00196f: @@ EmbeddedStub      909 73 84 dexp_stub (stub gen)
0x00007fd2dc002ed7: @@ EmbeddedStub      660 73 85 dlog_stub (stub gen)
0x00007fd2dc00316b: @@ EmbeddedStub      744 73 86 dlog10_stub (stub gen)
0x00007fd2dc001cfc: @@ EmbeddedStub      4571 73 87 dpow_stub (stub gen)
0x00007fd2dbfff84c: @@ EmbeddedStub      1861 73 88 dsin_stub (stub gen)
0x00007fd2dbffff91: @@ EmbeddedStub      1793 73 89 dcos_stub (stub gen)
0x00007fd2dc000692: @@ EmbeddedStub      2312 73 90 dtan_stub (stub gen)
0x00007fd2dc000f9a: @@ EmbeddedStub      1029 73 91 dsinh_stub (stub gen)
0x00007fd2dc00139f: @@ EmbeddedStub      859 73 92 dtanh_stub (stub gen)
0x00007fd2dc0016fa: @@ EmbeddedStub      629 73 93 dcbrt_stub (stub gen)
0x00007fd2dc003470: @@ EmbeddedStub      489 73 94 fmod_stub (stub gen)
0x00007fd2dbfff18e: @@ EmbeddedStub      1 73 95 verify_mxcsr_stub (stub gen)
0x00007fd2dbfff18f: @@ EmbeddedStub      56 73 96 f2i_fixup_stub (stub gen)
0x00007fd2dbfff1c7: @@ EmbeddedStub      67 73 97 f2l_fixup_stub (stub gen)
0x00007fd2dbfff20a: @@ EmbeddedStub      80 73 98 d2i_fixup_stub (stub gen)
0x00007fd2dbfff25a: @@ EmbeddedStub      90 73 99 d2l_fixup_stub (stub gen)
0x00007fd2dbfff2d0: @@ EmbeddedStub      16 73 100 float_sign_mask_stub (stub gen)
0x00007fd2dbfff2f0: @@ EmbeddedStub      16 73 101 float_sign_flip_stub (stub gen)
0x00007fd2dbfff310: @@ EmbeddedStub      16 73 102 double_sign_mask_stub (stub gen)
0x00007fd2dbfff330: @@ EmbeddedStub      16 73 103 double_sign_flip_stub (stub gen)
```

---------

---------
### Progress

### Issue
 * [JDK-8388410](https://bugs.openjdk.org/browse/JDK-8388410): [Leyden] Improve map file logging to add Embedded Stubs from StubGenBlobs (**Sub-task** - P4)

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - **Reviewer**)
 * [Andrew Dinn](https://openjdk.org/census#adinn) (@adinn - **Reviewer**)
 * [Ioi Lam](https://openjdk.org/census#iklam) (@iklam - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/118/head:pull/118` \
`$ git checkout pull/118`

`$ git checkout pull/118` \
`$ git pull https://git.openjdk.org/leyden.git pull/118/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 118`

`$ git pr show -t 118`

</details>

<a href="https://git.openjdk.org/leyden/pull/118.diff">https://git.openjdk.org/leyden/pull/118.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/118#issuecomment-4995658087)
</details>

### DIFF
--- a/src/hotspot/share/cds/aotMapLogger.cpp
+++ b/src/hotspot/share/cds/aotMapLogger.cpp
@@ -144,9 +144,9 @@ void AOTMapLogger::log_embedded_stubs(const AOTCodeCache* cache, const AOTCodeEn
     pos += sizeof(uint);
 
     // Log the embedded stub
-    log_debug(aot, map)(PTR_FORMAT ": @@ %-17s %d %d %d %s",
+    log_debug(aot, map)(PTR_FORMAT ": @@ %-17s %d id=%d blob=%d %s",
       p2i(buf + entry->offset() + entry->code_offset() + offset),
-      "EmbeddedStub", size, entry->id(), (int) stub_id, StubInfo::name(stub_id));
+      "EmbeddedStub", size, (int) stub_id, entry->id(), StubInfo::name(stub_id));
 
     //Get the number of secondary/extra entry offsets
     int n = *(int*) (buf + pos);
--- a/src/hotspot/share/cds/aotMapLogger.cpp
+++ b/src/hotspot/share/cds/aotMapLogger.cpp
@@ -123,31 +123,12 @@ void AOTMapLogger::dumptime_log(ArchiveBuilder* builder, FileMapInfo* mapinfo,
 void AOTMapLogger::log_embedded_stubs(const AOTCodeCache* cache, const AOTCodeEntry* entry) {
 
   //start at the beginning of our AOT cache
-  const char* buf  = cache->store_buffer();
+  const char* buf = cache->store_buffer();
 
-  //pos now points at the beginning of the StubGenBlob
-  uint pos = entry->offset() + entry->code_offset();
+  // jump to the embedded stubs section
+  uint pos = entry->offset() + entry->embedded_stub_offset();
 
-  // The first part is the CodeBlob which we are not going to log
-  const CodeBlob* blob = (const CodeBlob*)(buf + pos);
-  pos += blob->size();
-
-  // Then we have the relocation data records (aligned to HeapWordSize)
-  // we are not interested in logging this either
-  int reloc_count = *(int*)(buf + pos);
-  pos += sizeof(int);
-  pos = align_up(pos, (uint)HeapWordSize);
-  pos += reloc_count * (uint)sizeof(relocInfo);
-
-  // Then we have an optional OopMapSet
-  if (entry->has_oop_maps()) {
-    pos = align_up(pos, (uint)sizeof(int));
-    int oopmaps_size = *(int*)(buf + pos);  pos += sizeof(int);
-    pos += oopmaps_size;
-  }
-  pos = align_up(pos, (uint)sizeof(int));
-
-  // Now we can get the information we wanted to log (embedded stubs)
+  // Now we can get the information we want to log (embedded stubs)
   // get the first embedded stub id
   StubId stub_id = *(StubId*)(buf + pos);
   pos += sizeof(StubId);
--- a/src/hotspot/share/cds/aotMapLogger.cpp
+++ b/src/hotspot/share/cds/aotMapLogger.cpp
@@ -138,6 +138,8 @@ void AOTMapLogger::log_embedded_stubs(const AOTCodeCache* cache, const AOTCodeEn
 
   // loop until we run out of stubs (no stubid)
   while (stub_id != StubId::NO_STUBID) {
+    assert(stub_id > StubId::NO_STUBID && stub_id < StubId::NUM_STUBIDS, "Embedded Stub ID is not valid.");
+    assert(StubInfo::blob(stub_id) == entry->id(), "We found an embedded stub that doesn't belong here.");
     uint offset = *(uint*)(buf + pos);
     pos += sizeof(uint);
     uint size = *(uint*)(buf + pos);
@@ -153,6 +155,8 @@ void AOTMapLogger::log_embedded_stubs(const AOTCodeCache* cache, const AOTCodeEn
     pos += sizeof(int);
     //to skip them and prepare for the following stub (if exists)
     pos += n * sizeof(uint);
+    // the entry+extras count for the stub read from the file should exceed the declared entry count
+    assert(n >= StubInfo::entry_count(stub_id) - 1, "We are missing entries on this stub.");
 
     // position ourselves in the potential following stub
     stub_id = *(StubId*)(buf + pos);
--- a/src/hotspot/share/cds/aotMapLogger.cpp
+++ b/src/hotspot/share/cds/aotMapLogger.cpp
@@ -139,7 +139,7 @@ void AOTMapLogger::log_embedded_stubs(const AOTCodeCache* cache, const AOTCodeEn
   // loop until we run out of stubs (no stubid)
   while (stub_id != StubId::NO_STUBID) {
     assert(stub_id > StubId::NO_STUBID && stub_id < StubId::NUM_STUBIDS, "Embedded Stub ID is not valid.");
-    assert(StubInfo::blob(stub_id) == entry->id(), "We found an embedded stub that doesn't belong here.");
+    assert(StubInfo::blob(stub_id) == (BlobId) entry->id(), "We found an embedded stub that doesn't belong here.");
     uint offset = *(uint*)(buf + pos);
     pos += sizeof(uint);
     uint size = *(uint*)(buf + pos);
--- a/src/hotspot/share/cds/aotMapLogger.cpp
+++ b/src/hotspot/share/cds/aotMapLogger.cpp
@@ -145,7 +145,7 @@ void AOTMapLogger::log_ac_region() {
 
       switch (entry->kind()) {
         case  AOTCodeEntry::Kind::Nmethod:
-          log_debug(aot, map)(PTR_FORMAT ": @@ %-17s %d %d %d %s",
+          log_debug(aot, map)(PTR_FORMAT ": @@ %-17s %d level=%d id=%d %s",
             p2i(entry), AOTCodeCache::get_kind_name(entry->kind()),
             entry->size(), entry->comp_level(), entry->comp_id(), name);
           break;
@@ -154,16 +154,16 @@ void AOTMapLogger::log_ac_region() {
             BlobId blob_id = (BlobId) entry->id();
 
             //First we log the stub blob
-            log_debug(aot, map)(PTR_FORMAT ": @@ %-17s %d %d %s",
+            log_debug(aot, map)(PTR_FORMAT ": @@ %-17s %d id=%d %s",
               p2i(entry), AOTCodeCache::get_kind_name(entry->kind()),
               entry->size(), entry->id(), StubInfo::name(blob_id));
 
             //Now we log each stub embedded inside the blob
-            cache->loop_over_embedded_stubs(entry,  log_embedded_stub);
+            cache->iterate_embedded_stubs(entry,  log_embedded_stub);
           }
             break;
           default:
-          log_debug(aot, map)(PTR_FORMAT ": @@ %-17s %d %d %s",
+          log_debug(aot, map)(PTR_FORMAT ": @@ %-17s %d id=%d %s",
             p2i(entry), AOTCodeCache::get_kind_name(entry->kind()),
             entry->size(), entry->id(), name);
           break;
--- a/src/hotspot/share/cds/aotMapLogger.cpp
+++ b/src/hotspot/share/cds/aotMapLogger.cpp
@@ -120,6 +120,65 @@ void AOTMapLogger::dumptime_log(ArchiveBuilder* builder, FileMapInfo* mapinfo,
   log_info(aot, map)("[End of AOT cache map]");
 }
 
+void AOTMapLogger::log_embedded_stubs(const AOTCodeCache* cache, const AOTCodeEntry* entry) {
+
+  //start at the beginning of our AOT cache
+  const char* buf  = cache->store_buffer();
+
+  //pos now points at the beginning of the StubGenBlob
+  uint pos = entry->offset() + entry->code_offset();
+
+  // The first part is the CodeBlob which we are not going to log
+  const CodeBlob* blob = (const CodeBlob*)(buf + pos);
+  pos += blob->size();
+
+  // Then we have the relocation data records (aligned to HeapWordSize)
+  // we are not interested in logging this either
+  int reloc_count = *(int*)(buf + pos);
+  pos += sizeof(int);
+  pos = align_up(pos, (uint)HeapWordSize);
+  pos += reloc_count * (uint)sizeof(relocInfo);
+
+  // Then we have an optional OopMapSet
+  if (entry->has_oop_maps()) {
+    pos = align_up(pos, (uint)sizeof(int));
+    int oopmaps_size = *(int*)(buf + pos);  pos += sizeof(int);
+    pos += oopmaps_size;
+  }
+  pos = align_up(pos, (uint)sizeof(int));
+
+  // Now we can get the information we wanted to log (embedded stubs)
+  // get the first embedded stub id
+  StubId stub_id = *(StubId*)(buf + pos);
+  pos += sizeof(StubId);
+
+  // Embedded StubGenBlobs are assumed to have the following structure
+  // [ StubId | offset | size | N | offset_1 | offset_2 | ... | offset_N ]
+
+  // loop until we run out of stubs (no stubid)
+  while (stub_id != StubId::NO_STUBID) {
+    uint offset = *(uint*)(buf + pos);
+    pos += sizeof(uint);
+    uint size = *(uint*)(buf + pos);
+    pos += sizeof(uint);
+
+    // Log the embedded stub
+    log_debug(aot, map)(PTR_FORMAT ": @@ %-17s %d %d %d %s",
+      p2i(buf + entry->offset() + entry->code_offset() + offset),
+      "EmbeddedStub", size, entry->id(), (int) stub_id, StubInfo::name(stub_id));
+
+    //Get the number of secondary/extra entry offsets
+    int n = *(int*) (buf + pos);
+    pos += sizeof(int);
+    //to skip them and prepare for the following stub (if exists)
+    pos += n * sizeof(uint);
+
+    // position ourselves in the potential following stub
+    stub_id = *(StubId*)(buf + pos);
+    pos += sizeof(StubId);
+  }
+}
+
 void AOTMapLogger::log_ac_region() {
   ResourceMark rm;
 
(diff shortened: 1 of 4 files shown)


Delawen:
Windows test are failing but I am quite sure it is not my fault (??) It doesn't even finish the configuration.

https://github.com/Delawen/leyden/actions/runs/29523768519/job/87707205844

ashu-mehra:
> Windows test are failing but I am quite sure it is not my fault (??) It doesn't even finish the configuration.

If I remember correctly I have seen it failing in mainline as well. I don't think it is related to your changes.

src/hotspot/share/code/aotCodeCache.hpp
```diff
@@ -603,6 +603,7 @@ class AOTCodeCache : public CHeapObj<mtCode> {
   AOTCodeCache(bool is_dumping, bool is_using);
 
   const char* cache_buffer() const { return _load_buffer; }
+  const char* store_buffer()      const { return _store_buffer; }
```
vnkozlov: Please align (too many spaces)

vnkozlov (commented):
My concern is the way you find the offset of embedded stubs information is fragile.
May be we should consider adding an other offset field to AOTCodeEntry so you can directly use it. Or may be union with _comp_level and _comp_id and have add separate AOTCodeEntry() constructor for stubs.

Delawen:
> May be we should consider adding an other offset field to AOTCodeEntry so you can directly use it. Or may be union with _comp_level and _comp_id and have add separate AOTCodeEntry() constructor for stubs.

Do we need a separated AOTCodeEntry constructor for any other reason? If not, adding an extra offset field seems more clean to me.

iklam:
It's common for the existing logs to use the first number as the size. E.g.

```
0x00000008000d3788: @@ TypeArrayU8       96
0x00000008000d3788:   000000000000000b 000000080033a7e8 000000080033a730 000000080033a678   ..........3.....0.3.....x.3.....
0x00000008000d37a8:   000000080033a608 000000080033a598 000000080033a528 000000080033a4b8   ..3.......3.....(.3.......3.....
0x00000008000d37c8:   000000080033a448 000000080033a3d8 000000080033a368 000000080033a2f8   H.3.......3.....h.3.......3.....
0x00000008000d37e8: @@ TypeArrayU8       24
0x00000008000d37e8:   0000000000000002 00000008000dd000 00000008000dc800                    ........................
```

However, the meaning of the other number is not easy to understand. 
```
0x00007fd2fbffefe4: @@ StubGenBlob       24452 73 initial_blob (stub gen)
0x00007fd2dbfff05b: @@ EmbeddedStub      276 73 77 call_stub_stub (stub gen)
```

I think it's better to do this

```
0x00007fd2fbffefe4: @@ StubGenBlob       24452 id = 73 initial_blob (stub gen)
0x00007fd2dbfff05b: @@ EmbeddedStub      276 id = 77 blob = 73 call_stub_stub (stub gen)
```

(I swapped the order of `id` and `blob` so the `id` always appears first.

Also, is `(stub gen)` always printed?

adinn (approved):
All looks good.

vnkozlov:
> Do we need a separated AOTCodeEntry constructor for any other reason? If not, adding an extra offset field seems more clean to me.

I agree to new offset field.

vnkozlov:
Yes, having `(stab gen)` in all names is strange. @adinn do we really need it?

Delawen:
> Yes, having `(stab gen)` in all names is strange. @adinn do we really need it?

I just print what I get, if it is needed we can also scrap that before printing on the log (hope it won't break things further)

adinn:
> Yes, having (stub gen) in all names is strange. @adinn do we really need it?

Well, yes, we can avoid it but it is not as simple or as clean as it looks. Omitting the`(stub gen)` suffix has implications for other print contexts.

The need for the suffix is an unfortunate legacy of the way names have been given to stubs/blobs. We have several places where the names used to declare blobs and their associated stubs overlap across different groups of blob/stub.

So, for example,  `forward_exception` is used  declare a C1 blob
```
#define C1_STUBS_DO(do_blob)                                           \
  do_blob(dtrace_object_alloc)                                         \
  ...                                                                  \
  do_blob(forward_exception)                                           \
  ...                                                                  \
```
and also to declare an initial stubgen stub
```
#define STUBGEN_INITIAL_BLOBS_DO(do_blob, end_blob,                     \
                                 do_stub,                               \
                                 do_entry, do_entry_init,               \
                                 do_entry_array,                        \
                                 do_arch_blob,                          \
                                 do_arch_entry, do_arch_entry_init,     \
                                 do_arch_entry_array)                   \
  do_blob(initial)                                                      \
  do_stub(initial, call_stub)                                           \
  do_entry(initial, call_stub, call_stub_entry, call_stub_entry)        \
  ...                                                                   \
  do_stub(initial, forward_exception)                                   \
  do_entry(initial, forward_exception, forward_exception_entry,         \
           forward_exception_entry)                                     \
  ...                                                                   \
```

Another example is `new_instance` which is use in the declaration of both a C1 blob and its associated stub
```
#define C1_STUBS_DO(do_blob)                                           \
  do_blob(dtrace_object_alloc)                                         \
  ...                                                                  \
  do_blob(new_instance)                                                \
  ...                                                                  \
```
and a C2 blob and its associate stub
```
#define C2_STUBS_DO(do_blob, do_stub)                                  \
  do_blob(uncommon_trap, UncommonTrapBlob)                             \
  ...                                                                  \
  do_stub(new_instance, 0, true, false)                                \
  ...                                                                  \
```

Note that in the second example *in both cases* there is a C1 and C2 blob and an associated C1 and C2 stub which need to have a name generated for them both of them starting with the single declaration id `new_instance`

In order to ensure we have unique stub names (or unique blob names) we have to append the declared stub (or blob) name to the stub category name e.g. for the first case we generate stub names `"forward_exception (stub gen)"` and`"forward_exception (c1)"` and for the second case we generate stub (and blob) names `"new_instance (c1)"` and `"new_instance (c2)"`.

Note that in the first case the clash only occurs in the stub names. The blob names are `"initial (stub gen)"` and `"forward_exception (c1)"`. In the second case the clash occurs for both the blob name and the stub name (when we have a single stub blob we use the same name for the blob or stub).

So, even though the `(stub gen)` suffix appears to be redundant in the map printout -- because we only need to print the stub names for stubgen blobs i.e. multi-stub blobs -- it is actually an important element of the stub name that distinguishes it from other occurrences where a different stub is being named.

If we wanted to avoid having the suffix in the name then the way to achieve that would be to ensure that declaration names were unique across stub categories. We could, for example, just add a qualifying tag to the declaration name in the cases where names clash `forward_exception_stubgen` and `forward_exception_c1` -- likewise, `new_instance_c1` and `new_instance_c2`.

However, this would imply a change to the generated stub enum tags, field and accessor names for these cases requiring client code also to be updated. Calls to `SharedRuntime::forward_exception()` would now need to be written as `SharedRuntime::forward_exception_stubgen()`. Meanwhile other calls like `SharedRuntime::catch_exception()` would continue to omit the `_stubgen` suffix. Likewise, the code that accesses the C2 stub would now need to access `OptoRuntime::new_instance_c2_Java()`  while other calls like `OptoRuntime::new-array_Java()` would stay as they are.

I don't think this is actually much nicer as a solution. The map printout is slightly simplified but the client calls are now non-uniform.

adinn:
@vnkozlov I agree the current code that accesses the stub details is rather fragile -- I was thinking about that last night and thought of two ways to deal with it.

Firstly, we should probably move the code in print routine `AOTMapLogger::log_embedded_stubs` into class `AOTCodeCache`. That way if the layout gets changed then there is more chance that the print code will also get updated.

Secondly, we could usefully add some asserts to the printing code to ensure the data values read are as expected.

So, for example when we read the sizes for intermediate data to skip over the blob, the relocs and any oop map there should be asserts that the current position is within the bounds of the entry.

Likewise when reading successive stub ids
 - the stubid should be checked to ensure that is lies between `StubId::NO_STUBID` and `StubId::NUM_STUBIDS`
 - it should be an id associated with the current blob id `StubInfo::blob(stub_id) == blob_id`
 - the entry+extras count for the stub read from the file should exceed the declared entry count `count >= StubInfo::entry_count(blob_id)`

adinn (changes requested):
Maria, could you add asserts to the print code as mentioned in my comment to @vnkozlov and also consider moving the print routine into class AOTCodeCache?

Delawen:
Sure. I was already working on adding the offset as a field, I can do the rest too.

adinn:
@Delawen 
Just to confirm what we discussed online . . .

The entry + extras count for the stub that is read from the file always omits the first entry. Every stub must have at least one entry, at offset 0, and may have 0 or more extra entries. So, if you read value `count` then that value is supposed to represent `StubInfo::entry_count(blob_id) - 1 + extras`. So the minimum value for `count` is `StubInfo::entry_count(blob_id) - 1` and the asserted condition should be `count >= StubInfo::entry_count(blob_id) - 1`.

vnkozlov (approved):
Looks good to me.

vnkozlov:
@adinn 

I am okay with keeping `(stub gen)`. thank you for explaining the situation with it.

> Firstly, we should probably move the code in print routine AOTMapLogger::log_embedded_stubs into class AOTCodeCache. That way if the layout gets changed then there is more chance that the print code will also get updated.

Yes, I agree.

> Secondly, we could usefully add some asserts to the printing code to ensure the data values read are as expected.

Yes, we can add asserts. At lest check stubs ID.

src/hotspot/share/cds/aotMapLogger.cpp
```diff
@@ -120,6 +120,46 @@ void AOTMapLogger::dumptime_log(ArchiveBuilder* builder, FileMapInfo* mapinfo,
...
+    //Get the number of secondary/extra entry offsets
+    int n = *(int*) (buf + pos);
+    pos += sizeof(int);
+    //to skip them and prepare for the following stub (if exists)
+    pos += n * sizeof(uint);
+
+    // position ourselves in the potential following stub
+    stub_id = *(StubId*)(buf + pos);
+    pos += sizeof(StubId);
+  }
```
iklam: Do we have an iterator that can be used, instead of doing low level operations here in the AOT code?

For modularity, I would prefer an iterator that lives in the stub management code.

adinn: > Do we have an iterator that can be used, instead of doing low level operations here in the AOT code?

The nearest thing we have is class `AOTStubData`, which collects details of the stubs and associated entries as part of a traversal of the full entry contents. It supports full rehydration of the blob so it does a lot more work than we need simply to print details of the stubs.

Delawen: I tried using `AOTStubData` but I think it is overkilling. This check on `base` returns always the nullpointer: https://github.com/openjdk/leyden/blob/premain2/src/hotspot/share/code/aotCodeCache.cpp#L4235-L4238

Which leads me to think we can only use `AOTStubData` if/when we are loading the cache into memory, not when we are generating it. Modifying it further to be able to have a valid `StubAddrRange` seems too complex.

The code I'm using is https://github.com/Delawen/leyden/blob/tiriri/src/hotspot/share/cds/aotMapLogger.cpp#L125-L153

Delawen: So unless you have any other ideas I think this PR is ready to go.

adinn: Yes, AOTStubData is built to do a much more complex job than you need.

> So unless you have any other ideas I think this PR is ready to go.

You can still relocate the code that iterates over the stub entries in method `log_embedded_stub` from class `AOTMapLogger` into class `AOTCodeCcache`. You will need to refactor it a bit -- just transplanting the method as is moves the printing logic out of class `AOTMapLogger`. You can avoid that by implementing a method on AOTCodeCache that accepts a StubGen entry and a function pointer or a closure and applies the function to each stub in the entry e.g.

Declare a fn pointer type in `aotCodeCache.hpp`
```
typedef void (*stub_printer_t)(BlobId blob_id, StubId stub_id, int base, int code_base, int offset);
```
Then add a public method to class `AOTCodeCache`:

```
void AOTCodeCache::do_print_stub_details(AOTCodeEntry* entry, aot_stub_printer_t printer) {
  assert(entry->kind() == StubGen), "invalid call");
  BlobId blob_id = (BlobId)entry->id();
  int base = entry->offset();
  int code_base = entry->code_offset();
  //skip forward to entries
  . . .
   while (stub_id != StubId::NO_STUBID) {
     offset = ...;
     (*printer)(blob_id, stub_id, base, code_base, offset);
     // skip to next id
    . . .
  }
}
```
You can then call `AOTCodeCache::do_print_stub_details` from `AOTMapLogger` passing in a pointer to the print function. 

An alternative to using a fn pointer is to use a lambda or a closure i.e. an instance of a class that implements callback method.

Delawen: Pushed it http://github.com/openjdk/leyden/pull/118/commits/13276190295fb3922dd1e93fee903aa577504d04

But I am not sure what the NOT_CDS_RETURN or CDS_ONLY macros mean. I guess I should add them to my new functions?

src/hotspot/share/code/aotCodeCache.hpp
```diff
@@ -107,6 +107,7 @@ class AOTCodeEntry {
   uint   _name_offset; // Method's or intrinsic name
   uint   _name_size;
   uint   _code_offset; // Start of code in cache
+  uint   _embedded_stub_offset; // Start of embedded stubs in StubGenBlobs
```
adinn: I know @vnkozlov recommended insering this field but I don't really like it. It means every `AOTCodeEntry` now includes and extra `int` field even thought the offset is only needed for StubGen stub entries and then only when we are dumping a map.

The point of adding this field is to avoid having to traverse a StubGen entry's blob, relocs and oop map in order to locate the stub data. That was maybe a good idea when the code was outside of class `AOTCodeCache` as it reduced fragility. However, now that the traversal code is internal to the class (and now also that we have asserts to discover any errors introduced by a change in format) I think it is a better trade-off to omit this extra field and require method `loop_over_embedded_stubs` to traverse from the start of the entry to the start of the stub data. We will only do that traversal several hundreds of times when we are dumping a map but we will save an extra word for several thousands of nmethods in every cache.

@vnkozlov What do you think?

adinn (commented):
Looks good apart from a question as to whether we want to preserve the offset to the stub data which I am hoping @vnkozlov will address.

vnkozlov:
I like field. How about union?:

```
   union {
     uint   _embedded_stub_offset; // Start of embedded stubs in StubGenBlobs

     uint   _comp_level;  // compilation level
   };
```
You have to check `kind` in constructor to assign field to avoid overwriting.

Delawen:
Seems to work. @vnkozlov @adinn

adinn (approved):
Ok, looks good.

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -432,6 +432,48 @@ void AOTCodeCache::dump() {
   }
 }
 
+//This is a helper function that executes a function over each embedded stub available
+void AOTCodeCache::loop_over_embedded_stubs(const AOTCodeEntry* entry, const embedded_stub_func func) {
```
iklam: The HotSpot convention is to call this function `iterate_embedded_stubs()`.

src/hotspot/share/cds/aotMapLogger.cpp
```diff
@@ -143,7 +149,20 @@ void AOTMapLogger::log_ac_region() {
             p2i(entry), AOTCodeCache::get_kind_name(entry->kind()),
             entry->size(), entry->comp_level(), entry->comp_id(), name);
           break;
-        default:
+        case  AOTCodeEntry::Kind::StubGenBlob:
+          {
+            BlobId blob_id = (BlobId) entry->id();
+
+            //First we log the stub blob
+            log_debug(aot, map)(PTR_FORMAT ": @@ %-17s %d %d %s",
```
iklam: For consistency, use `"... %-17s %d id=%d %s"`,

Also, we should use `id=%d` and `level=%d` for the `Nmethod` and `default` cases.

src/hotspot/share/code/aotCodeCache.hpp
```diff
@@ -160,6 +168,9 @@ class AOTCodeEntry {
   uint name_size()    const { return _name_size; }
   uint code_offset()  const { return _code_offset; }
 
+  void set_embedded_stub_offset(const uint off) { _embedded_stub_offset = off; }
+  uint embedded_stub_offset()  const { return _embedded_stub_offset; }
```
iklam: These functions should have `precond(_kind == StubGenBlo)`. Similarly, the `comp_level()` accessor should assert the opposite.

src/hotspot/share/code/aotCodeCache.hpp
```diff
@@ -137,7 +140,12 @@ class AOTCodeEntry {
     _name_size    = name_size;
     _code_offset  = code_offset;
 
-    _comp_level   = comp_level;
+    if (_kind == StubGenBlob) {
+      _embedded_stub_offset = 0;
+    } else {
+      _comp_level   = comp_level;
```
iklam: Extraneous spaces before `=`.

vnkozlov:
Looks good to me too. I will wait GHA finished to submit our testing too.

Delawen:
@vnkozlov ready for your tests

iklam (approved):
LGTM

vnkozlov:
> @vnkozlov ready for your tests

Submitted.

vnkozlov (approved):
My testing passed.

adinn (approved):
good

Delawen:
Hope the failing windows test job does not interfere with integration...

Delawen:
Can someone sponsor me? @adinn @ashu-mehra @iklam @vnkozlov

PR #128, opened 2026-08-16
8390443: [leyden] AOTCodeEntry::_not_entrant should be synchronized

vnkozlov:
I am not sure about these changes. @iklam, @veresov, @adinn and @ashu-mehra please help.

At first `AOTCodeCache::invalidate(AOTCodeEntry)` was only called from `nmethod::make_not_entrant()` under NMethodState_lock. So changing `_not_entrant field` was synchronized.

Later two additional places were added: `MethodCounters::deallocate_contents()` which was called during class unloading and `MethodCounters::clear_counters()` called from WhiteBox API `clearMethodState()`.

These changes added new `AOTCodeCache::invalidate_with_lock()` method which takes `NMethodState_lock`. It will be used in those two methods.

On other hand these 2 methods may never be called for methods with AOT code. `deallocate_contents()` is not called for methods from built-in loaders but it may not be true for future when we add support for custom class loaders (do we plan to generate AOT **preload** code for them). `clear_counters()` -  we don't support WhiteBox API for AOT code currently but it may change in a future.

---------

---------
### Progress

### Issue
 * [JDK-8390443](https://bugs.openjdk.org/browse/JDK-8390443): [leyden] AOTCodeEntry::_not_entrant should be synchronized (**Enhancement** - P4)

### Reviewers
 * [Ioi Lam](https://openjdk.org/census#iklam) (@iklam - **Reviewer**)
 * [Ashutosh Mehra](https://openjdk.org/census#asmehra) (@ashu-mehra - **Reviewer**)
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/128/head:pull/128` \
`$ git checkout pull/128`

`$ git checkout pull/128` \
`$ git pull https://git.openjdk.org/leyden.git pull/128/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 128`

`$ git pr show -t 128`

</details>

<a href="https://git.openjdk.org/leyden/pull/128.diff">https://git.openjdk.org/leyden/pull/128.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/128#issuecomment-5309069881)
</details>

### DIFF
--- a/src/hotspot/share/code/aotCodeCache.cpp
+++ b/src/hotspot/share/code/aotCodeCache.cpp
@@ -603,19 +603,8 @@ AOTCodeCache::AOTCodeCache(bool is_dumping, bool is_using) :
 
 void AOTCodeCache::invalidate(AOTCodeEntry* entry) {
   // Invalidate AOT code during production run.
-  // This should be called only from nmethod::make_not_entrant().
-  assert_lock_strong(NMethodState_lock);
+  // This could be concurrent execution.
   if (entry != nullptr && is_on_for_use()) {
-    _cache->invalidate_entry(entry);
-  }
-}
-
-void AOTCodeCache::invalidate_with_lock(AOTCodeEntry* entry) {
-  // Invalidate AOT code during production run.
-  // This could be concurrently executed - use lock.
-  if (entry != nullptr && is_on_for_use()) {
-    ConditionalMutexLocker ml(NMethodState_lock, !NMethodState_lock->owned_by_self(),
-                              Mutex::_no_safepoint_check_flag);
     _cache->invalidate_entry(entry);
   }
 }
@@ -1098,6 +1087,19 @@ void* AOTCodeEntry::operator new(size_t x, AOTCodeCache* cache) {
   return (void*)(cache->add_entry());
 }
 
+bool AOTCodeEntry::not_entrant() const {
+  return AtomicAccess::load(&_not_entrant);
+}
+
+void AOTCodeEntry::set_not_entrant() {
+  return AtomicAccess::store(&_not_entrant, true);
+}
+
+bool AOTCodeEntry::try_set_not_entrant() {
+  // If old value is 'true' - other thread set it already.
+  return AtomicAccess::xchg(&_not_entrant, true);
+}
+
 static bool check_entry(AOTCodeEntry::Kind kind, uint id, uint comp_level, AOTCodeEntry* entry) {
   if (entry->kind() == kind) {
     assert(entry->id() == id, "sanity");
@@ -1169,7 +1171,7 @@ AOTCodeEntry* AOTCodeCache::find_entry(AOTCodeEntry::Kind kind, uint id, uint co
 
 void AOTCodeCache::invalidate_entry(AOTCodeEntry* entry) {
   assert(entry!= nullptr, "all entries should be read already");
-  if (entry->not_entrant()) {
+  if (entry->try_set_not_entrant()) {
     return; // Someone invalidated it already
   }
 #ifdef ASSERT
@@ -1197,7 +1199,6 @@ void AOTCodeCache::invalidate_entry(AOTCodeEntry* entry) {
   found = (i < count);
   assert(found, "entry should exist");
 #endif
-  entry->set_not_entrant();
   uint name_offset = entry->offset() + entry->name_offset();
   const char* name = _load_buffer + name_offset;;
   uint level       = entry->comp_level();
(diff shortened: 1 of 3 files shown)


vnkozlov:
Looking through pass to `MethodCounters::deallocate_contents()` and it seems safe. It is executed only at safe point. And based on next comment AOT nmethod should be "deoptimized" already and aot code entry marked as `not_entrant`:
```
// Release Method*.  The nmethod will be gone when we get here because
// we've walked the code cache.
void Method::deallocate_contents(ClassLoaderData* loader_data) {
```

iklam:
> Looking through pass to `MethodCounters::deallocate_contents()` and it seems safe. It is executed only at safe point. And based on next comment AOT nmethod should be "deoptimized" already and aot code entry marked as `not_entrant`:
> 
> ```
> // Release Method*.  The nmethod will be gone when we get here because
> // we've walked the code cache.
> void Method::deallocate_contents(ClassLoaderData* loader_data) {
> ```

Maybe we don't need to grab the lock, but in `AOTCodeCache::invalidate()`:

```
 assert_locked_or_safepoint(NMethodState_lock);
```

ashu-mehra:
For custom loaders if and when we generate the preload code, we should probably avoid marking the `AOTCodeEntry` as invalid unless each classloader has its own AOTCodeEntry. If a CL1 loads A::foo() and then later marks its AOTCodeEntry invalid, then another loader CL2 that loads A::foo() won't be able to load the entry.

ashu-mehra:
In current code how is the synchronization between a thread T1 invalidating an AOTCodeEntry and another thread T2 trying to use that entry achieved? T1 will be holding `NMethodState_lock`. Does T2 also acquire `NMethodState_lock` when loading the aot code entry?

vnkozlov:
> For custom loaders if and when we generate the preload code, we should probably avoid marking the `AOTCodeEntry` as invalid unless each classloader has its own AOTCodeEntry. If a CL1 loads A::foo() and then later marks its AOTCodeEntry invalid, then another loader CL2 that loads A::foo() won't be able to load the entry.

I think we should have separate AOT code and corresponding entry for each class loader.

vnkozlov:
> In current code how is the synchronization between a thread T1 invalidating an AOTCodeEntry and another thread T2 trying to use that entry achieved? T1 will be holding `NMethodState_lock`. Does T2 also acquire `NMethodState_lock` when loading the aot code entry?

Invalidation is done only for loaded/installed AOT code with `AOTCodeEntry::_loaded` field set to true during AOT code installation into CodeCache. The field never change its state after that.

When T2 tries to find AOT code it calls `find_aot_code_entry()` which checks both `_loaded` and `_not_entrant` fields. And only when both fields are negative it returns entry: [compilationPolicy.cpp#L94](https://github.com/openjdk/leyden/blob/premain2/src/hotspot/share/compiler/compilationPolicy.cpp#L94)

There is `_not_entrant` field check in `check_entry()` in aotCodeCache.cpp but it is only called from `AOTCodeCache::find_code_entry()` for nmethods.

ashu-mehra:
> When T2 tries to find AOT code it calls find_aot_code_entry() which checks both _loaded and _not_entrant fields. And only when both fields are negative it returns entry: [compilationPolicy.cpp#L94](https://github.com/openjdk/leyden/blob/premain2/src/hotspot/share/compiler/compilationPolicy.cpp#L94)

Right, I missed the check for `_loaded`.

ashu-mehra:
I think you are right. `nmethod` is already purged and freed by the time `MethodCounters::deallocate_contents` is called. So it should be safe to call `AOTCodeCache::invalidate()` without lock. btw why do we even care to synchronize the access to `_not_entrant` during invalidation?

vnkozlov:
> btw why do we even care to synchronize the access to `_not_entrant` during invalidation?

That is why I asked your opinion about these changes in Description. 

The only thing we do is set `_not_entrant` to `true` and then may invalidate entry for preload AOT code (set their field too). It actually  does not matter which thread (or few threads) sets this field.  The only critical read of this field is in  `find_aot_code_entry()` and is guarded by `!is_loaded()` check.  The only drawback is possible duplicated logs output.

There are other `_not_entrant`  reads in `AOTCodeCache::finish_write()` but they are redundant (leftover from development). The invalidation is guarded by `is_on_for_use()` - we don't invalidate entries during AOT code compilation (assembly phase). I need to clean them up (or replace with asserts).

May be I should update comment in `AOTCodeCache::invalidate()` to add details so that AI is not freaking out about it:
```
  // Invalidate AOT code during production run.
  // This could be concurrent execution.
```

vnkozlov:
I can use `AtomicAccess::xchg()` in `AOTCodeCache::invalidate_entry()` and atomic loads in other places to avoid new locks.  May be this is preferable?

ashu-mehra:
> The only critical read of this field is in find_aot_code_entry() and is guarded by !is_loaded() check.

I agree. From functional correctness the code looks fine. `!is_loaded()` check prevents reusing an entry being invalidated. To prevent duplicate log messages, I think using `AtomicAccess::xchg` should be sufficient.

vnkozlov:
Thank you @ashu-mehra. I will work on that.

vnkozlov:
I updated changes with atomic access. I also cleaned up `finish_write()` code which had left over from early development.

iklam (approved):

ashu-mehra (approved):
lgtm

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -1110,6 +1087,19 @@ void* AOTCodeEntry::operator new(size_t x, AOTCodeCache* cache) {
...
+  return AtomicAccess::load(&_not_entrant);
+}
+
+void AOTCodeEntry::set_not_entrant() {
+  return AtomicAccess::store(&_not_entrant, true);
+}
+
+bool AOTCodeEntry::try_set_not_entrant() {
+  // If old value is 'true' - other thread set it already.
+  return AtomicAccess::xchg(&_not_entrant, true);
```
shipilev, line 1100: So, should `_not_entrant` be `Atomic<bool>` then, so that we gain access to its instance methods? Also, for these concurrent tries, we usually do `return _not_entrant.compare_set(false, true)`, and return `true` when try was succeeded. It inverts the meaning that you have now.

shipilev (approved):
Looks okay, with nits

src/hotspot/share/code/aotCodeCache.cpp
vnkozlov: Thank you, @shipilev, for suggestion. I am not familiar with `Atomic<bool>` API. I think can make it with current code:
```
  bool AOTCodeEntry::try_set_not_entrant() {
    // If old value is 'true' - other thread set it already.
    return AtomicAccess::xchg(&_not_entrant, true) == false;
  }
  ...
      if (!entry->try_set_not_entrant()) {
      return; // Another thread already invalidated it.
```

vnkozlov: On other hand I like current code with comments which explains what happened.

vnkozlov:
Thank you all for reviews and suggestions.

PR #134, opened 2026-09-08
8391565: [leyden] AOT C1 compilation should load null reset value from class

adinn:
This fixes AOT C1 compilation of a `putfield` to a flat inline field by calling `null_reset_value` at runtime to retrieve the prototype null value rather than calling it at assembly time and hard-wiring the oop it into the compiled method's constant table and code.

---------

---------
### Progress

### Issue
 * [JDK-8391565](https://bugs.openjdk.org/browse/JDK-8391565): [leyden] AOT C1 compilation should load null reset value from class (**Bug** - P4)

### Reviewers
 * [Vladimir Kozlov](https://openjdk.org/census#kvn) (@vnkozlov - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/134/head:pull/134` \
`$ git checkout pull/134`

`$ git checkout pull/134` \
`$ git pull https://git.openjdk.org/leyden.git pull/134/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 134`

`$ git pr show -t 134`

</details>

<a href="https://git.openjdk.org/leyden/pull/134.diff">https://git.openjdk.org/leyden/pull/134.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/134#issuecomment-5585754739)
</details>

### DIFF
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -1827,6 +1827,16 @@ void GraphBuilder::copy_inline_content(ciInlineKlass* vk, Value src, int src_off
   }
 }
 
+Value GraphBuilder::load_null_reset_value(ciInlineKlass* vk) {
+  if (AOTCodeCache::is_dumping_code()) {
+    Value clazz = append(new Constant(new ClassConstant(vk)));
+    Value offset = append(new Constant(new IntConstant(vk->get_null_reset_value_offset())));
+    return append(new UnsafeGet(T_OBJECT, clazz, offset, /*is_volatile*/false,/*is_raw*/true));
+  } else {
+    return append(new Constant(new ObjectConstant(vk->get_null_reset_value().as_object())));
+  }
+}
+
 void GraphBuilder::access_field(Bytecodes::Code code) {
   bool will_link;
   ciField* field = stream()->get_field(will_link);
@@ -2142,7 +2152,8 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
 
           // Store the subfields when field is a nullable non-atomic field
           Value object_null = append(new Constant(objectNull));
-          Value null_reset_value = append(new Constant(new ObjectConstant(inline_klass->get_null_reset_value().as_object())));
+          // Value null_reset_value = append(new Constant(new ObjectConstant(inline_klass->get_null_reset_value().as_object())));
+          Value null_reset_value = load_null_reset_value(inline_klass);
           Value src = append(new IfOp(val, Instruction::neq, object_null, val, null_reset_value, state_before, false));
           copy_inline_content(inline_klass, src, inline_klass->payload_offset(), obj, offset, state_before);
 
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -1828,7 +1828,7 @@ void GraphBuilder::copy_inline_content(ciInlineKlass* vk, Value src, int src_off
 }
 
 Value GraphBuilder::load_null_reset_value(ciInlineKlass* vk) {
-  if (AOTCodeCache::is_dumping_code()) {
+  if (compilation()->env()->is_aot_compile()) {
     Value clazz = append(new Constant(new ClassConstant(vk)));
     Value offset = append(new Constant(new IntConstant(vk->get_null_reset_value_offset())));
     return append(new UnsafeGet(T_OBJECT, clazz, offset, /*is_volatile*/false,/*is_raw*/true));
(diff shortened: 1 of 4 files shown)


adinn:
With this in place (and with AOT code caching enabled when Valhalla is enabled) it is possible to build an AOT cache for the test program mentioned in the issue. However, a production run using the cache falls over running AP 4 compiled code for method `LocalDate.create()`. The AP compilation inlines a terminal call to 'new LocalDate()' and ends up returning the result by value. This involves caching the 3 field values in `r5`-`r7` and returning an inline oop 'buffer' in `r0`. The buffer appears to be allocated during C2 compilation and has (slightly weird) address `0x3f800137c01`.

src/hotspot/share/c1/c1_GraphBuilder.cpp
```diff
@@ -1827,6 +1827,16 @@ void GraphBuilder::copy_inline_content(ciInlineKlass* vk, Value src, int src_off
   }
 }
 
+Value GraphBuilder::load_null_reset_value(ciInlineKlass* vk) {
+  if (AOTCodeCache::is_dumping_code()) {
```
vnkozlov: Please use: `compilation()->env()->is_aot_compile()`

src/hotspot/share/c1/c1_GraphBuilder.cpp
```diff
@@ -2142,7 +2152,8 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
 
           // Store the subfields when field is a nullable non-atomic field
           Value object_null = append(new Constant(objectNull));
-          Value null_reset_value = append(new Constant(new ObjectConstant(inline_klass->get_null_reset_value().as_object())));
+          // Value null_reset_value = append(new Constant(new ObjectConstant(inline_klass->get_null_reset_value().as_object())));
```
vnkozlov: Left over?

src/hotspot/share/ci/ciInlineKlass.cpp
```diff
@@ -166,6 +166,12 @@ ciConstant ciInlineKlass::get_null_reset_value() {
   return ciConstant(T_OBJECT, CURRENT_ENV->get_object(null_reset_value));
 }
 
+int ciInlineKlass::get_null_reset_value_offset() {
+  VM_ENTRY_MARK
+  InlineKlass* vk = get_InlineKlass();
+  return vk->null_reset_value_offset();
```
vnkozlov, line 172: Is this constant between runs? You use `IntConstant` for it which will be embedded into code.

vnkozlov:
Please, file bug on C2 if you did not.

src/hotspot/share/ci/ciInlineKlass.cpp
adinn: Yes, it is constant and will be retained from assembly through to production:

1. `_null_reset_value_offset` is an offset into the static field data located in the mirror of the `InlineKlass`. It identifies a hidden field (named using `vmSymbols::null_reset_value_name()`). Changing the offset between runs would require the static field layout algorithm to vary.
2. The hidden field is one of three possible hidden fields injected into an `InlineKlass` under `ClassFileParser::parse_fields` during class file parsing (the field offset is recorded during a subsequent walk of the field sequence by class `FieldLayoutBuilder`). The hidden fields always follows all the declared static fields plus any static fields returned by `JavaClasses::get_injected()`. Which of these three extra `InlineKlass` fields is included depends on the class definition but for any given class it will always be the same fields.
3. The field is located in the `Members` instance that appears at the end of every `InlineKlass` directly following all the data that belongs to its `InstanceKlass` segment. The size of an `InlineKlass` is computed by adding in the size of this `Members` data to the size of the `InstanceKlass` segment. So, when the `InstanceKlass` is saved to cache and later restored this `Members` data area and all its content will be saved and restored automatically, *modulo* removal and recreation of unshareable content. The current code that removes and restores unshareable content does not appear to modify any fields of the `Members` segment.

src/hotspot/share/c1/c1_GraphBuilder.cpp
adinn: Done

src/hotspot/share/c1/c1_GraphBuilder.cpp
adinn: Yes, removed now.

adinn:
I raised https://bugs.openjdk.org/browse/JDK-8392086 for the buffer constant problem.

src/hotspot/share/ci/ciInlineKlass.cpp
vnkozlov: Good. Thank you for detailed investigation.

vnkozlov (approved):
Looks good.

PR #132, opened 2026-08-27
8391294: [leyden] Add support for AOT linking of classes loaded by URLClassLoader

ashu-mehra:
Following description is generated with the help of AI to aid reviewers in following the code.

## Overview

This commit adds support for caching classes loaded by `java.net.URLClassLoader` instances in the AOT (Ahead-Of-Time) cache. Previously, only classes loaded by the three built-in loaders (boot, platform, app) could be AOT-linked. Classes loaded by any other ("custom") loader were treated as "unregistered" -- they could be archived but not AOT-linked, so they couldn't benefit from pre-resolved constant pool entries or pre-linked class hierarchies.

This change introduces the concept of an **"AOT-safe" custom loader** -- a `URLClassLoader` whose parent is a built-in loader and whose classpath consists entirely of JAR files. For these loaders, the full AOT-linking machinery (pre-resolved CP entries, archived module/package metadata, archived `ProtectionDomain` objects) is extended to work end-to-end.

---

## Three-Phase Lifecycle

The AOT cache workflow has three phases. The change touches all three:

```
Training Run (incl. recipe recording) ──> Assembly (apply recipes) ──> Production Run
```

### Phase 1: Training Run

**Entry point:** `URLClassLoader` constructor (Java side)

Every `URLClassLoader` constructor now calls `registerForAOTLinking(urls)`.

**Flow:**

```
URLClassLoader.<init>(urls, parent)
  └─ registerForAOTLinking(urls)
       └─ isSafeForAOTLinking(urls)   -- checks:
       │    (a) `this.getClass() == URLClassLoader.class` (not a subclass)
       │    (b) parent is boot/platform/system loader
       │    (c) all URLs are file:// protocol ending in .jar
       └─ registerForAOTLinkingImpl(parent, classpath)    [native]
            └─ JVM_RegisterURLClassLoaderForAOTLinking()   [jvm.cpp]
```

**JVM-side registration (`jvm.cpp`):**

1. **Compute a deterministic AOT identity** by concatenating `<parent_aot_id><path_separator><classpath>`. For example, if the parent is the system loader (identity `"SYSTEM"`) and the classpath is `/app/lib.jar`, the identity becomes `"SYSTEM:/app/lib.jar"`. This identity is the key that ties the training-run loader to the assembly-phase and production-run loaders.

2. **Reserve the identity** in `ClassLoaderAotIdTable` (atomically via mutex) to prevent duplicate registrations.

[...]

PR #135, opened 2026-09-11
8391294: [leyden] Add support for AOT linking of classes loaded by URLClassLoader

ashu-mehra:
This PR adds support for AOT loading and linking of classes loaded by `java.net.URLClassLoader` instances. 

In the current state of AOTCache only the 3 built-in loaders and their classes benefit from full AOT linking and constant pool resolution. Classes loaded by custom loaders are considered "unregistered" and are not preloaded/prelinked. 

This PR introduces the notion of aot-safe custom loaders. Classes loaded by such loaders get the same treatment as the 3 built-in loaders. Currently only the direct instances of `URLClassLoader` which have jar files on disk as their classpath are considered aot-safe.

**Training phase**

During training phase when a `URLClassLoader` instance is created, JVM performs some checks to determine if it is safe to be considered for AOTCache optimizations. This is done in `URLClassLoader.isSafeForAOTLinking()`. If the checks pass, then it is assigned an ID which comprises of its parent's classpath and its own classpath (classpath is created by concatenating all the URLs). This ID is stored in `ClassLoaderAotIdTable` which is a map of ID -> ClassLoaderData (see `JVM_RegisterURLClassLoaderForAOTLinking()`). 

When a class is loaded by this classloader, JVM sets a bit `defined_by_aot_safe_custom_loader` in `InstanceKlass::_misc_flags`. It also records the index in the classpath from where this class has been loaded (see `ClassLoader::record_result()`). The index is used in the assembly phase for caching the `ProtectionDomain`, `URL` and `Manifest` objects corresponding to this location (similar to `CDSProtectionDomain`).

When the training run ends, `FinalImageRecipes` (see `record_aot_safe_loader_classes()`) records the list of all aot-safe custom loaders and their classes in the preimage (or the configuration file). Data structure for recording custom loader recipes is `ClassLoaderRecipe` in `FinalImageRecipes.hpp`. It also records array of `AOTClassLocation` corresponding to the classpath of the URLClassLoader. This is required to recreate scratch loader during the assembly phase.

**Assembly phase**

In the assembly phase `FinalImageRecipes` iterates through `ClassLoaderRecipe` list and preloads and prelinks all the classes (`FinalImageRecipes::load_aot_safe_custom_loader_classes()`). It first verifies that the files referred by the classpath have not been modified on disk, and then creates a scratch `URLClassLoader` instance using the classpath recorded in the `ClassLoaderRecipe` which is then used to load the classes. The loading is done by `SystemDictionary::load_class_from_preimage()` which is similar to `SystemDictionary::preload_class()` used for pre-loading classes during production run. Before initiating the loading, the assembly phase creates the `ProtectionDomain`, `URL` and `Manifest` objects and caches them in `KlassMirrorDataCache` similar to the way `CDSProtectionDomain` is doing.

At the end of assembly phase when generating the AOTCache, JVM creates a `CustomLoaderInfo` structure for each aot-safe custom loader to store its metadata, like its aot id, classpath, PackageEntry table, ModuleEntry table along with a table of classes loaded by the loader. This data structure is used in the production phase to preload the classes when a new `URLClassLoader` instance is created. These structures are stored in `ArchivedCustomLoaderInfoMap` which is a map of ID -> CustomLoaderInfo.

Note that at this point the `ProtectionDomain` and `Module` objects in the archived heap have references to the scratch `URLClassLoader` instances.

**Production phase**

When a `URLClassLoader` instance is created, JVM checks `ArchivedCustomLoaderInfoMap` to find `CustomLoaderInfo` corresponding to the loader's ID. If a match is found, it validates the classpath of the loader with the archived classpath in `CustomLoaderInfo`. If that succeeds, it proceeds to preload all the classes stored in `CustomLoaderInfo` (`AOTLinkedClassBulkLoader::preload_classes_for_loader()`). During preloading, it patches the `ProtectionDomian` and `Module` objects in the archived heap to point to the new `URLClassLoader` instance being created. The patching is done in `SystemDictionary::preload_class()` and `ModuleEntry::restore_archived_oops()`. It also marks the class loader as the initiating loader of all the parent's classes.

TODO:
1. This patch does not handle `Class-Path Attribute` in the jar's manifest file.
2. Handling of custom "jar" protocol handler 

---------

---------
### Progress
- \[ \] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))

### Issue
 * [JDK-8391294](https://bugs.openjdk.org/browse/JDK-8391294): [leyden] Add support for AOT linking of classes loaded by URLClassLoader (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/135/head:pull/135` \
`$ git checkout pull/135`

`$ git checkout pull/135` \
`$ git pull https://git.openjdk.org/leyden.git pull/135/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 135`

`$ git pr show -t 135`

</details>

<a href="https://git.openjdk.org/leyden/pull/135.diff">https://git.openjdk.org/leyden/pull/135.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/135#issuecomment-5628293675)
</details>


src/java.base/share/classes/java/net/URLClassLoader.java
```diff
@@ -92,6 +97,7 @@ public class URLClassLoader extends SecureClassLoader implements Closeable {
     public URLClassLoader(URL[] urls, ClassLoader parent) {
         super(parent);
         this.ucp = new URLClassPath(urls);
+        registerForAOTLinking();
```
ashu-mehra, line 100: https://github.com/openjdk/leyden/pull/132#discussion_r3875385178

> UCP does a defensive copy of the URL[], this is what will need to checked to see if it can be registered as you can't use the mutable array that the caller has a reference to.

Updated `registerForAOTLinking()` to use `ucp.getURLs()`.

src/java.base/share/classes/java/net/URLClassLoader.java
```diff
@@ -639,4 +649,87 @@ public static URLClassLoader newInstance(final URL[] urls) {
...
+
+    /**
+     * Returns true if the parent is a builtin loader
+     *
+     * @return true if the parent is a builtin loader
+     */
+    @SuppressWarnings("this-escape")
+    private final boolean hasBuiltinLoaderAsParent() {
+        ClassLoader parent = getParent();
+        return parent == null || parent == ClassLoaders.platformClassLoader() || parent == ClassLoaders.appClassLoader();
```
ashu-mehra, line 661: https://github.com/openjdk/leyden/pull/132#discussion_r3875367959

> The system class loader is configurable, it may not be built-in. The application class loader is built-in and I think this is what you need here.

Updated the code to use `ClassLoaders.appClassLoader()` instead of `getSystemClassLoader()`.

src/java.base/share/classes/java/net/URLClassLoader.java
```diff
@@ -639,4 +649,87 @@ public static URLClassLoader newInstance(final URL[] urls) {
...
+                if (file.startsWith("file:") && file.endsWith("!/")) {
+                    // strip off "!/"
+                    @SuppressWarnings("deprecation")
+                    URL nestedURL = new URL(file.substring(0, file.length() - 2));
+                    url = nestedURL;
+                } else {
+                    return false;
+                }
+            }
+            if (!urlToCheck.getProtocol().equals("file") || !urlToCheck.getPath().endsWith(".jar")) {
```
ashu-mehra, line 686: https://github.com/openjdk/leyden/pull/132#discussion_r3875357875
> URI scheme are compared without regard to case.

It appears the protocol is converted to lower case in URL constructor. https://github.com/openjdk/jdk/blob/e20054271886ef6c6649d10824444a3e5dc53e3a/src/java.base/share/classes/java/net/URL.java#L482
So the upper case check is not really needed.

src/java.base/share/classes/java/net/URLClassLoader.java
```diff
@@ -639,4 +649,87 @@ public static URLClassLoader newInstance(final URL[] urls) {
...
+                if (file.startsWith("file:") && file.endsWith("!/")) {
+                    // strip off "!/"
+                    @SuppressWarnings("deprecation")
+                    URL nestedURL = new URL(file.substring(0, file.length() - 2));
+                    url = nestedURL;
+                } else {
+                    throw new AssertionError("Unexpected url " + url);
+                }
+           }
+           String path = Path.of(url.toURI()).toString();
```
ashu-mehra, line 709: https://github.com/openjdk/leyden/pull/132#discussion_r3875442906

> This is a URL path component so cannot be used as a file path, e.g./d/foo%20bar.jar, so will need a decoding step. There's an API note in the URL API docs on this.

Updated to use `Path.of(URI)` instead of `url.getPath()`.

ashu-mehra:
I have addressed the review comments mentioned in earlier PR #132 in this PR.

Ioi mentioned here https://github.com/openjdk/leyden/pull/132#discussion_r3876576461

> It's better to put this field into RunTimeClassInfo. It's more cumbersome, but will not add a footprint penalty for every dynamically loaded class.

I managed to remove to need to store classloader's aot id in InstanceKlass or RunTimeClassInfo. It was mainly used to check if the class has been loaded by aot-safe loader. I achieved the same by setting a bit in `InstanceKlass._misc_flags`.

src/hotspot/share/cds/aotLinkedClassBulkLoader.hpp
```diff
@@ -76,6 +78,9 @@ class AOTLinkedClassBulkLoader :  AllStatic {
   static void replay_training_at_init_for_preloaded_classes(TRAPS) NOT_CDS_RETURN;
 
   static bool is_initializing_classes_early() NOT_DEBUG({return false;});
+
+  static void preload_classes_for_loader(ClassLoaderData* loader_data, CustomLoaderInfo* cl_info, TRAPS);
+  static void link_classes_for_loader(ClassLoaderData* loader_data, CustomLoaderInfo* cl_info, TRAPS);
```
adinn, line 83: These might be better renamed to `preload_classes_for_custom_loader` and `link_classes_for_custom_loader` in order to clarify their use.

src/hotspot/share/cds/aotMetaspace.cpp
```diff
@@ -1116,11 +1126,10 @@ void AOTMetaspace::load_classes(TRAPS) {
 
 void AOTMetaspace::exercise_runtime_cds_code(TRAPS) {
   // Exercise the manifest processing code
```
adinn:
```suggestion
  // Exercise Java manifest processing code using
  // a dummy jar and manifest, forcing the associated
  // Java classes to be loaded into the current VM.
```

src/hotspot/share/cds/archiveUtils.hpp
```diff
@@ -280,20 +280,20 @@ class ReadClosure : public SerializeClosure {
 };
 
 class ArchiveUtils {
-  template <typename T> static Array<T>* archive_non_ptr_array(GrowableArray<T>* tmp_array);
-  template <typename T> static Array<T>* archive_ptr_array(GrowableArray<T>* tmp_array);
+  template <typename T> static Array<T>* archive_non_ptr_array(GrowableArrayView<T>* tmp_array);
```
adinn: Perhaps the template specifications for this method and the next one ought to include, respectively, `ENABLE_IF(!std::is_pointer<T>::value)>`  and `ENABLE_IF(std::is_pointer<T>::value)>`. There is no error here because they are only called from the appropriate variants of `archive_array` which specify the same condition guards but 1) someone could get that wrong later and 2) specifying the extra template condition reinforces the intended usage. Note that doing that would also require a corresponding addition to the template /definitions/ provided in `archiveUtils.inline.hpp` but they would need to employ macro `ENABLE_IF_SDEFN`.

src/hotspot/share/cds/cdsConfig.cpp
```diff
@@ -664,6 +665,10 @@ bool CDSConfig::check_vm_args_consistency(bool patch_mod_javabase, bool mode_fla
   if (AOTClassLinking) {
     // If AOTClassLinking is specified, enable all these optimizations by default.
     FLAG_SET_ERGO_IF_DEFAULT(AOTInvokeDynamicLinking, true);
+    FLAG_SET_ERGO_IF_DEFAULT(AOTCacheSupportForCustomLoader, true);
```
adinn: Are there no circumstances where we might want to reject a user setting for `AOTCacheSupportForCustomLoader` in the else branch?

PR #138, opened 2026-09-21
8392086: [leyden] AOT C2 compilation fails for methods that return scalarized values

vnkozlov:
After fixing the issue in C2 my testing found several typos and incorrect code in previous Valhalla related fixes. I fixed them too. Now I can run `aotCode/` tests with `--enable-preview` (after clearing problem list and commenting Valhalla check in AOTCodeCache).  I added regression test based on the test from @adinn.

 I did not enable AOT code caching with Valhalla because I need to fix and other small issue: [JDK-8390239](https://bugs.openjdk.org/browse/JDK-8390239). Which I decided to do separately.

Tested tier1-3 with AOT code enabled with Valhalla.

---------

---------
### Progress

### Issue
 * [JDK-8392086](https://bugs.openjdk.org/browse/JDK-8392086): [leyden] AOT C2 compilation fails for methods that return scalarized values (**Bug** - P4)

### Reviewers
 * [Andrew Dinn](https://openjdk.org/census#adinn) (@adinn - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/138/head:pull/138` \
`$ git checkout pull/138`

`$ git checkout pull/138` \
`$ git pull https://git.openjdk.org/leyden.git pull/138/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 138`

`$ git pr show -t 138`

</details>

<a href="https://git.openjdk.org/leyden/pull/138.diff">https://git.openjdk.org/leyden/pull/138.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/leyden/pull/138#issuecomment-5754768042)
</details>


src/hotspot/share/opto/valuetypenode.cpp
```diff
@@ -1743,6 +1743,14 @@ Node* ValueTypeNode::is_loaded(PhaseGVN* phase, ciValueKlass* vk, Node* base, in
 
 Node* ValueTypeNode::tagged_klass(ciValueKlass* vk, PhaseGVN& gvn) {
   const TypeKlassPtr* tk = TypeKlassPtr::make(vk);
+#if INCLUDE_CDS
+  if (gvn.C->env()->is_aot_compile()) {
+    // Keep klass as metadata to make it relocatable for AOT code.
+    Node* k_con = gvn.makecon(tk);
+    Node* bits  = gvn.transform(new CastP2XNode(nullptr, k_con));
+    return gvn.transform(new OrLNode(bits, gvn.longcon(1)));
```
adinn, line 1751: I'm not really clear how this works but I searched through the opto code and it looks like we can be sure that this subgraph `(OrL (CastP2X ConP[tk]) 1)` is not going to be reduced to a simple `ConL` because that only happens when the `ConP` has type `RawPtr` and in this case it has type `KlassPtr`.
Is that correct? Or is there some other reason we can be sure the OR of two constants does not get folded into a single constant?

adinn (approved):
Apologies for the generated code errors. I got it to compile but was not able to run it. All looks good now.

src/hotspot/share/opto/valuetypenode.cpp
vnkozlov: makecon(tk) creates metadata type constant keeping its type (connote.cpp:60).
CastX2P only converts Raw pointer type constants to raw bits (castnode.cpp:646).
As result metadata type constant will not be folded. And .ad will use metadata_type relocation because constant has metadata type:
```
0x00007f9d3bee8625:   mov    $0x1,%eax
0x00007f9d3bee862a:   movabs $0x9e189000,%r10             ;   {metadata('java/time/LocalDate')}
0x00007f9d3bee8634:   or     %r10,%rax
```

vnkozlov:
Thank you. @adinn, for review.

PR #139, opened 2026-09-21
8392746: [Leyden] Add missing GC barrier relocation support to AOT code

vnkozlov:
AOT code relocation processing is missing relocInfo::barrier_type. It is used by ZGC. After adding it I found missing relocation in Aarch64 and RISC Z GC specific code.

It was fixed in AOT code PR already. Porting it to leyden:premain2 branch.

Testing: tier1-3

---------

---------
### Progress
- \[ \] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))

### Issue
 * [JDK-8392746](https://bugs.openjdk.org/browse/JDK-8392746): [Leyden] Add missing GC barrier relocation support to AOT code (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/leyden.git pull/139/head:pull/139` \
`$ git checkout pull/139`

`$ git checkout pull/139` \
`$ git pull https://git.openjdk.org/leyden.git pull/139/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 139`

`$ git pr show -t 139`

</details>

<a href="https://git.openjdk.org/leyden/pull/139.diff">https://git.openjdk.org/leyden/pull/139.diff</a>

</details>


vnkozlov:
@adinn please look

src/hotspot/share/code/aotCodeCache.cpp
```diff
@@ -2608,6 +2608,15 @@ bool AOTCodeCache::write_relocations(CodeBlob& code_blob, RelocIterator& iter,
         reloc_data.at_put(idx, delta);
         break;
       }
+      case relocInfo::barrier_type: {
+        // Patched by GC during nmethod registration
+        assert(code_blob.is_nmethod(), "Only nmethod has barrier relocation");
```
vnkozlov, line 2614: I left assert to catch possible future issues.

shipilev:
In recent mainline, there is also `patchable_barrier_type`, used by Shenandoah. It carries more metadata, but otherwise it is similar to existing one. You can just add it to the same `case` block.

vnkozlov:
@shipilev, thank you for pointing this to me. Do we need to preserve/restore metadata in AOT code cache? I see it has `uint16_t _metadata;` field.

And is it also used only in `nmethod`? So I can keep current `nmethod` checks.

vnkozlov:
I added `patchable_barrier_type`. I ran locally on linux-x64 `aotCode` tests with `-XX:+UseShenandoahGC` and they passed. My local JDK build includes  Shenandoah GC.

shipilev:
> @shipilev, thank you for pointing this to me. Do we need to preserve/restore metadata in AOT code cache? I see it has `uint16_t _metadata;` field.

Yes, we better save the metadata as well. That metadata includes the critical pieces of info GC needs.

> And is it also used only in `nmethod`? So I can keep current `nmethod` checks.

Yes, only in nmethods.

adinn (approved):

adinn:
I tagged the two locations in the JEP PR where the fallthrough for `relocInfo::patchable_barrier_type` is needed.

vnkozlov:
Thank you, @adinn

